PR #557, opened 2014-12-20
Persistent cache 2

mvglasow:
Includes improvements discussed with Emux and Ludwig:
- Read entire cache dir when initializing a persistent cache
- Reliably enforce cache size for persistent cache (data from previous instances is now included in the count)
- Fixed NPE which occasionally occurred in FileSystemTileCache#containsKey() when persistence was enabled
- Added test case for persistent tile cache
- Added missing compatibility constructors and helper methods for apps migrating from 0.5.0
- Use file timestamp to determine age of map file
- Changed TTL to TimeToLive in identifiers
- Comment for default TTL (86400000 millis = 1 day)
- Changed default Mapnik TTL to 8279s (current OSM setting) and added comment

As mentioned before, in order to enforce cache size (and also to fix the NPE Ludwig found) I now examine the entire cache dir when the cache is initialized and enter anything I find there in the lruCache. This might affect startup performance if the cache dir holds a large number of files.

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/util/AndroidUtil.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/util/AndroidUtil.java
@@ -47,11 +47,12 @@ import android.view.WindowManager;
 public final class AndroidUtil {
 
 	public static final boolean HONEYCOMB_PLUS = Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB;
+	
 	/**
 	 * Creates a two-level tile cache.
 	 * <p>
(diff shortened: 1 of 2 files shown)


mvglasow:
Hi, You're not gonna believe what I've read yesterday, you've got to read it yourself at <http://thujugyxo.jacobharry.com/aedjbz>

Sincerely, Michael von Glasow

mvglasow:
Hello friend, 

I've  found that  store  with a lot  of amazing new stuff,  just take a look, you'll  love it <http://verify.abbyposner.net/aesjyc>

Thanks for your consideration, Peter von Glasow

PR #563, opened 2015-01-06

kcwu:
Fix formating and bugs in doc.

Note that the sign bit of VBE-S is in last byte, not first byte.


devemux86:
Thanks @kcwu

PR #564, opened 2015-01-06
Rename incorrect variable name enlargementInPixel

kcwu:
What the code doing is enlarging by meter, not by pixel


devemux86:
You are right, these ```LatLongUtils``` methods have meters as parameters.

applantation:
Yes thanks, keep this stuff coming. It is good to fix the small things.

PR #599, opened 2015-04-09
AssetsRenderTheme shouldn't ignore relativePathPrefix

lincomatic:
AssetsRenderTheme was ignoring relativePathPrefix, so symbol & pattern folders had to be in root of assets.

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
@@ -43,7 +43,6 @@ public class AssetsRenderTheme implements XmlRenderTheme {
 	public AssetsRenderTheme(Context context, String relativePathPrefix, String fileName, XmlRenderThemeMenuCallback menuCallback) throws IOException {
 		this.assetName = fileName;
 		this.relativePathPrefix = relativePathPrefix;
-        this.inputStream = context.getAssets().open((this.relativePathPrefix == null) ? "" : this.relativePathPrefix + this.assetName);
 		this.menuCallback = menuCallback;
 	}
 
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
@@ -43,7 +43,7 @@ public class AssetsRenderTheme implements XmlRenderTheme {
 	public AssetsRenderTheme(Context context, String relativePathPrefix, String fileName, XmlRenderThemeMenuCallback menuCallback) throws IOException {
 		this.assetName = fileName;
 		this.relativePathPrefix = relativePathPrefix;
-		this.inputStream = context.getAssets().open(this.assetName);
+        this.inputStream = context.getAssets().open((this.relativePathPrefix == null) ? "" : this.relativePathPrefix + this.assetName);
 		this.menuCallback = menuCallback;
 	}
 
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
@@ -43,6 +43,7 @@ public class AssetsRenderTheme implements XmlRenderTheme {
 	public AssetsRenderTheme(Context context, String relativePathPrefix, String fileName, XmlRenderThemeMenuCallback menuCallback) throws IOException {
 		this.assetName = fileName;
 		this.relativePathPrefix = relativePathPrefix;
+		this.inputStream = context.getAssets().open((this.relativePathPrefix == null) ? "" : this.relativePathPrefix + this.assetName);
 		this.menuCallback = menuCallback;
 	}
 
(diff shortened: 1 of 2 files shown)


devemux86:
Thanks for the PR.

Actually the ```relativePathPrefix``` was primarily introduced for the case we use a render theme in assets that retrieve its resources from a jar (like mapsforge jar).
So the ```relativePathPrefix``` was the relative path inside the jar itself.

But I understand that we may have a similar case with assets paths too.

Nevertheless the ```relativePathPrefix``` is about the resources only.
The theme path can be fully described with the ```fileName``` parameter.
So I don't think we need the change in ```AssetsRenderTheme``` class.

devemux86:
A general note, please try following our code and style conventions (e.g. indentation).

devemux86:
We can make rendertheme-v4 external by replacing the ```assets:``` with ```file:``` prefix.
You just need to have in the same dir the render theme xml and the patterns / symbols folders, a common way of packaging render themes.

The ```relativePathPrefix``` is not ignored, as said it's being handled when we use an assets render theme that declares resources from a jar.

About your PR, can you repost it with the ```AndroidGraphicFactory``` changes?

lincomatic:
Actually, you just described the exact reason for my PR. Without my PR, when I use rendertheme-v4 as file: or jar: theme, it looks for the patterns/symbols folders as subfolder of relativePathPrefix, so I can package the whole theme in its own folder. But when used as assets:, it ignores relativePathPrefix, so the symbols/patterns are always relative to the root of assets.

Sorry, I am not very god with git. I don't know why the commit deleted my change to AndroidGraphicFactory, rather than fixing the indent. Will resubmit now.

devemux86:
Yes the ```relativePathPrefix``` is needed for assets too, the ```AndroidGraphicFactory``` changes are useful.

About GitHub, it's recommended for every PR to create a (dev) topic branch in your repository.
That allows you to build new features or test out ideas without putting your main beanch at risk.
Of course you need to rebase it each time dev changes.
(e.g. I have updated dev from your last fork)

You can read more in [GitHub Help](https://help.github.com/articles/fork-a-repo/).

lincomatic:
That is great advice. Thank you for the tip.

devemux86:
@lincomatic your latest code seems ok, but we don't need the changes for ```AssetsRenderTheme```.
It refers to render theme xml path, which is described fine with the ```fileName``` parameter.

We only need the changes in ```AndroidGraphicFactory``` for the resources themselves.

lincomatic:
Sorry for taking so long to reply.. I didn't see your latest comment until now.
Actually, we also need the change in AssetsRenderTheme, because the user instantiates the rendertheme by calling the constructor of AssetsRenderTheme(). If we only use the changes in AndroidGraphicFactory, there's no way to pass the relativePathPrefix down to AndroidGraphicFactory, so the change is useless.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
```diff
@@ -43,7 +43,7 @@ public AssetsRenderTheme(Context context, String relativePathPrefix, String file
 	public AssetsRenderTheme(Context context, String relativePathPrefix, String fileName, XmlRenderThemeMenuCallback menuCallback) throws IOException {
 		this.assetName = fileName;
 		this.relativePathPrefix = relativePathPrefix;
```
devemux86: The ```relativePathPrefix``` is passed down here.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/AssetsRenderTheme.java
```diff
@@ -43,7 +43,7 @@ public AssetsRenderTheme(Context context, String relativePathPrefix, String file
...
-		this.inputStream = context.getAssets().open(this.assetName);
+		this.inputStream = context.getAssets().open((this.relativePathPrefix == null) ? "" : this.relativePathPrefix + this.assetName);
```
devemux86: Here we parse the render theme xml file, i.e. ```fileName``` parameter which became ```assetName``` variable.

The file path can be totally described by this, so we don't need any mix with the ```relativePathPrefix``` which is for the resources paths in the xml itself.

devemux86:
Any news on this?

lincomatic:
Excluding this change:
-		this.inputStream = context.getAssets().open(this.assetName);
+		this.inputStream = context.getAssets().open((this.relativePathPrefix == null) ? "" : this.relativePathPrefix + this.assetName);

forces the theme XML file to be stored in the root of the assets, unless we instantiate AssetsRenderTheme with fileName=relativePathPrefix+fileName. I have a rendertheme that's stored like this:
osma/
osma/mysomatheme.xml
osma/patterns
osma/symbols

Without my change, we need to use
AssetsRenderTheme(context,"osma","osma/mysosmatheme.xml"....
This is redundant, and unexpected, since the parameter is called "fileName," not filePath. 
If you feel strongly about it, I don't mind, It might be easier to just close this PR and make the changes the way you like them, since my PR ugly, with 3 commits.

devemux86:
Ok now I understand better what you mean.
I don't oppose to the changes, but can you squash the commits into one?

I posted some instructions here:
https://groups.google.com/d/msg/mapsforge-dev/AK3yjjd8E6s/a6Ct5V7b-i8J

lincomatic:
I tried squashing but every time I synced it, the squash would get undone, because I had already pushed it to the remote. Hope you don't mind.. I just created a new PR instead. 
https://github.com/mapsforge/mapsforge/pull/612

devemux86:
It's ok, Git is very powerful but it can get equally hard to manage too.

PR #612, opened 2015-05-19
AssetsRenderTheme shouldn't ignore relativePathPrefix

lincomatic:
see https://github.com/mapsforge/mapsforge/pull/599


devemux86:
Thanks @lincomatic for the catch and the PR.

(I just improved a bit the implementation)

PR #648, opened 2015-07-14
Feature: public map read result builder

muZZkat:
I need to write a custom MapDataStore so needed to create make the MapReadResult builder public.

I left the variables with no modifier so they could still be added directly by the package, I guess it would be better to mark them provide and make the package also access them via the getter/settings.

Did you want me to make this change?

### DIFF
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
@@ -563,7 +563,7 @@ public class MapFile implements MapDataStore {
 
 		// the query is finished, was the water flag set for all blocks?
 		if (queryIsWater && queryReadWaterInfo) {
-			mapReadResultBuilder.isWater = true;
+			mapReadResultBuilder.setWater(true);
 		}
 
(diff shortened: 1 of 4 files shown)


devemux86:
You mean ```private```, right?

>Did you want me to make this change?
Yes please do.

muZZkat:
`private`, not `provide` :)

ludwigb:
+1 for this change

devemux86:
Also we can rename the ```setIsWater``` setter to ```setWater```, (variable's name confuses the generators).

muZZkat:
OK. Should all be sweet.

devemux86:
You did the rename via refactoring?
The ```setWater``` should be changed in all needed files.

Also could you squash the commits into one?
You can find instructions [here](https://groups.google.com/d/msg/mapsforge-dev/AK3yjjd8E6s/a6Ct5V7b-i8J), or I can do it if you find it difficult.

muZZkat:
Never done that before. Learn something new every day. Commits have been squashed.

devemux86:
We try to keep repository clean with the final commits. :)

Thanks for the PR.

PR #677, opened 2015-09-16
Fix LatLongUtils.zoomForBounds() can return negative values, #678

schildbach:
I got -1 from that method so I decided to fix the sanity check.


devemux86:
I see we already perform value restriction, with min=0.
So we need to check why -1 managed to be returned.

schildbach:
Math.min doesn't work because it doesn't catch wrapovers from the byte conversion.

devemux86:
Do you have an example (to include also in test class) ?

devemux86:
BTW let's open an issue for this, and then solve it.
That way people would know what's fixed.

schildbach:
E.g. if the result of Math.max(...) would be 255, the narrowed byte value would be -1.

devemux86:
I see what you mean, that could be solved by casting ```(byte)``` the ```floor``` result, but if/then works too.

devemux86:
Fixes #678.

PR #686, opened 2015-09-17
Setup Travis continuous integration.

schildbach:
To enable it, all the mapsforge account owner needs to do it log in on https://travis-ci.org and flip the enable switch for the mapsforge repository.

In a future update, we could add other forms of notifications. Currently, it notifies via GitHub (very convenient!) and I think email.

A preview can be seen on my account: https://travis-ci.org/schildbach/mapsforge
If you want to try out how pull requests work, just open a PR against my GitHub repo and you'll see the Travis integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+# configuration for https://travis-ci.org/mapsforge/mapsforge
+sudo: false
+language: android
+jdk: openjdk7
+android:
+  components:
+    - build-tools-23.0.1
+    - android-22
+script: gradle build


devemux86:
We'd like to see if we need to include another CI system (besides Jenkins) in our list.

Examining the Travis file, does it manage that Mapsforge is a multi-language project (Android, Java) with branches based on different Android SDK versions (22, 23) ?

schildbach:
I didn't even know Jenkins exists, because it didn't amend the GitHub pull requests. Anyway, as you can see Travis is extremely easy to set up and in my experience works much more reliable than Travis. The downside (or upside, depending on your priorities) is it's mainly a hosted service.

If a branch uses a different Android SDK you need to change the "- android-22" line on that branch as well.
If you need both SDKs on the same branch then you can also include a second line "- android-23", so you would have both available.

applantation:
Yes, why should we do this? What is the advantage over jenkins?

I hate to say this, but I do not even have the time to look at travis.

I am not principally opposed, but just no time....

devemux86:
[Travis](https://en.wikipedia.org/wiki/Travis_CI) is a hosted CI attached to GitHub and can help on PR test quality.
We can examine it but in time and via testing it.

So I'd propose to close the PR for now.
We need to check it accordingly first, probably putting our own settings in the process.

schildbach:
This PR is agnostic of the specific GitHub repository, as long as it's a fork of mapsforge. It works *as is* on my repo, as it would on any of yours, including the official one. So you can try it on a personal GitHub repo if you want. Just flip the switch on Travis.

schildbach:
Thanks for merging. I'm sure it will make developer live easier.

devemux86:
:+1:

ludwigb:
Just playing with it, setup was easy enough.

Does this give us downloadable libraries somewhere like we have on
ci.mapsforge.org for all our builds and releases?

devemux86:
I think we should add also ```oraclejdk7```.

schildbach:
@ludwigb I believe yes, but it needs to be configured. I can have a look at it if you want. E.g. for tags I'd try this setup: http://docs.travis-ci.com/user/deployment/releases/

devemux86:
Also in ```language``` put also ```java```.
And set ```android-23``` as dev is based on that.

schildbach:
You can only have one language. "android" is a superset of" java".

schildbach:
The sample app has android:targetSdkVersion="22" in its AndroidManifest.xml.

devemux86:
Target was set that way to avoid the new permissions model.
Compile version for dev is 23 and master 22.

schildbach:
I believe we can add oraclejdk7, then it will build on both JDKs.

schildbach:
According to Google, you can't avoid the new permission system like that. But anyway, I'll update travis.

devemux86:
>According to Google, you can't avoid the new permission system like that.

Can you post a link?
We thought differently: "Only apps designed to work with the Android M SDK (API level 23) will use the new permission system. App developers can stick to the old system, as long as they target API level 22 (Android 5.1) or earlier."

We just postpone the Samples modification for checking permissions in code.
(conversation [here](https://groups.google.com/forum/#!topic/mapsforge-dev/ySEKXU0NIKY))

schildbach:
I could not find the reference to Marshmallow. I believe the project.properties files are not used any more for gradle builds. In fact, if the build would reference Marshmellow it should fail, but it doesn't.

(It's still good to have the project.properties files for Eclipse though.)

schildbach:
@devemux86 Link: http://de.droidcon.com/session/whats-new-android
(I hope I get it right. It's what I remember from the talk.)

devemux86:
..yes not all like AS with Gradle..
So we need to support both IDE / build environments.

@schildbach thanks!
BTW I found also the reference [link](http://www.androidauthority.com/hiroshi-locheimer-permissions-android-m-616324/).

schildbach:
Yes, I think I was remembering this: "But what happens to older apps, designed for Android pre-M? This video from Google I/O explains that users will still be able to revoke permissions of these apps in Android M. In this case, apps will be fed blank data, which means they won’t break, and instead they will just show empty objects…"

ludwigb:
If this is how it works I personally cannot wait for Android M. Sick and
tired of apps that hoover down everything they can find on my phone.

devemux86:
:smile: 
Only concern is to not fill the code with continuous checks that hit performance (which I think are asynchronous too)..
Imagine Mapsforge having to check all the time if write permissions to cache have been revoked with the app paused/resumed.

schildbach:
Luckily you don't need permission to use *your* external cache directory from Android 4.4 onwards…

PR #694, opened 2015-09-18
Adjust default horizontal margin of controls a little.

schildbach:


devemux86:
I prefer keeping the existing minimal margins.
We have implemented setter methods for users to modify the margin.

schildbach:
Currently, it has a big vertical margin and only a small horizontal margin. I just wanted to make sure the defaults look good. We could also reduce the vertical margins if you prefer.

devemux86:
Which of the two components is asymmetrical and at what dpi?

At base 160dpi I see both scale bar and zoom controls with equal distances to the edges.
Maybe we miss a scale multiplication on high dpi?

devemux86:
Vertical margin is already 0, and horizontal was selected to be as small as possible and simultaneously symmetrical.

schildbach:
Here is a screenshot. My phone is xhdpi (as are most phones today). As you can see, there is quite a bit of vertical margin.

device-2015-09-18-203442

schildbach:
Oh, the screenshot revealed there is a copyright line at the bottom of the screen. Sorry, that line is ridiculously small, I thought it's a graphical glitch. Anyway, that line is perhaps causing the unwanted vertical spacing.

devemux86:
> My phone is xhdpi (as are most phones today)

I tend to disagree here.. :smile: 

Nevertheless it would never be perfect at all dpi, as zoom controls have their own vertical margin and scale bar one is dependent to font metrics too.

devemux86:
No the line is irrelevant (it's through layout system).
Again all these are to be customized per user.

schildbach:
Ok I agree my PR is probably only an inadequate fix. I'll try something more elaborate.

schildbach:
Customization is no excuse for a bad default look and feel!

devemux86:
We put them both as close to edges as possible. Then comes customization.
And "bad" is so subjective..

devemux86:
And please being firmly critical on an open source project does not appear nice to other people having invested valuable time creating and maintaining it..

devemux86:
And please being firmly judgemental on an open source project does not appear nice to other people having invested valuable time creating and maintaining it...

applantation:
The ridiculously small copyright line is to comply with OSM attribution ...
it does not say how big it has to be ....

schildbach:
@devemux86 Yes, I'm a bit picky when it comes to UI (-:  Don't take it personally. I try to apply criticism only when I can show it can be done better with a patch.

Issue #680, opened 2015-09-16
MapView as a layout for native Android views?


schildbach:
The old Google Maps v1 library had this wonderful feature where MapView would act as a geo-coordinate-aware layout. You would add views to the layout and they would follow all map panning and zooming.

```
mapView.addView(view, new MapView.LayoutParams(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT,
	point, 
MapView.LayoutParams.BOTTOM_CENTER));
```

Here is how the view could look. Other than markers etc. it contains native Android buttons.

![device-2015-09-16-234321](https://cloud.githubusercontent.com/assets/743306/9919378/bd50f29e-5ccc-11e5-9c26-1ed50aaa01f4.png)

Is something like this planned for mapsforge?

devemux86:
Agreed that it would be a useful feature to have in Mapsforge.
I have tried it in the past, not without issues at pan / zoom and neglected it.

Experiments could start as always in Samples app.

schildbach:
This is the next feature I will work on shortly, if no one else beats me at it.

devemux86:
Since we're dealing with a vector library, on top of my list would be allowing steplessly variable zoom levels (old issue #75). :smile:

devemux86:
Thanks, merged with some improvements. Some remarks:
* View overlays are not handled properly during zoom scale animation
* Osmdroid similar implementation has also additional x/y offsets for the alignment locations (might be unnecessary)

schildbach:
Thanks for merging.
Zoom scale animations: this will be fixed once we do proper zoom animations. Currently zoom animations are very local to MapViewPosition. Also see my comment about that on #688.
Offsets: if there is a usecase, we can easily add that. I never had one, as I was always anchoring to one of the 9 standard alignment points.

PR #695, opened 2015-09-19
Customizable zoom controls.

schildbach:
It now allows for vertical arrangements and customization of the button drawables. Gravity can now use any of the Gravity constants. Also, zoom speed of the buttons is configurable.

The samples are extended to show modern vertically arranged zoom controls and a screen density-aware position of those controls.

It now uses standard framework mechanisms to layout the controls. For this, MapActivity had to be switched to subclass FrameLayout rather than just ViewGroup.

Feel free to give it a whirl in the samples and let me know what you think.

device-2015-09-19-132202


devemux86:
Thanks will check it.

Though moving ```MapView``` from ```ViewGroup``` to ```FrameLayout``` is not something desirable.

We're going away from something generic (breaking users experience). There has to be a way to use ```ViewGroup``` for hosting like works now.

devemux86:
BTW a convenient way for custom zoom controls (and our advice to users) was always hosting ```MapView``` in another layout and add on top the controls with the desired placement.

schildbach:
Well, of course you can implement a ViewGroup too, but you'd end up duplicating FrameLayout. Why not use the Android framework?

devemux86:
Because it's more limiting and just for one use and specially for components which cannot be included in the library anyway (so few will use it).

Imagine if someone else may need it as ```RelativeLayout```.
And in #680 we may change our opinion again.

And to prevent reactions, *Mapsforge intention is not becoming Android Library or AAR*, we prefer our current jar universal state.

schildbach:
I just pushed a change that adds a convenient setControlOrientation() for the four possible combinations of orientation and which of the two controls comes first.

devemux86:
A friendly reminder, it's better [discussing](https://github.com/mapsforge/mapsforge/blob/master/docs/Getting-Started-Developers.md#how-to-contribute) first the development of new features or changes.

We appreciate your effort, and so it will not be spent on things that eventually we may be unable to apply.

schildbach:
I appreciate the reminder. However, I need those thing I implement anyway. The reason I'm using pull requests is because mapsforge is free software, and I thought it's only fair to contribute back.

schildbach:
Ok, I will try to use ViewGroup rather than FrameLayout.

devemux86:
>However, I need those thing I implement anyway.

That's understandable, but you need to see also that we have to keep Mapsforge as generic as possible for all users out there and more importantly not breaking their current API experience.

>The reason I'm using pull requests is because mapsforge is free software, and I thought it's only fair to contribute back.

Please keep them coming!

devemux86:
(Analyzing the PR)
* Let's keep existing ```MapZoomControls``` class intact without altering its structure with foreign interferences or hacks (also there are many using it *as is*).
* We can have a new class (e.g. ```CustomZoomControls```) with the proposed functionality (still playing with current ```MapView```).

schildbach:
I'm wondering will you actually post your analysis?

schildbach:
I just reworked this PR so that MapView is still a ViewGroup (rather than FrameLayout) and rebased on current dev.

devemux86:
Thanks.
* We can move customizable zoom controls into their own class (leaving default *as is*) and keeping only their code
* Use an ```enum``` way switching among those similar components
* Preferably the layout should be handled by the children, like was already implemented (we'll see for #699)

devemux86:
Merged 004dff4ea9133e8b5d4c899946941893dae06b0d.

schildbach:
Great, thanks for merging!

ludwigb:
Should we not also have drawables for other densities than hdpi?

devemux86:
>Should we not also have drawables for other densities than hdpi?

Yes having zoom drawables >```hdpi``` would be useful nowadays.

IIRC these drawables exist only in ```hdpi``` ?
(they have also their horizontal equivalents)

schildbach:
@ludwigb Since it's only horizontal and vertical lines it doesn't matter much. Also, it's only a sample. (The other samples use hdpi only as well.)

applantation:
I thought you cared about individual pixels? 😉.

Ok, maybe no issue.

schildbach:
:smirk:

Issue #696, opened 2015-09-19
Overlay/Controls enhancements


schildbach:
I will work on a few enhancements of the various controls and overlays, mostly look & feel.

First step is PR #695.

Next up: scale overlay, copyright overlay.

devemux86:
Same recommendation applies with scale bar, as I did with zoom controls at https://github.com/mapsforge/mapsforge/pull/695#issuecomment-141816438.

Keep current scale bar intact and have a new one with the proposed functionality (specially on heavy modifications).

After all we should support (and we are) multiple expressions of a map control.
That way new custom controls could have their own life and evolution.

devemux86:
Created #700 for customizable zoom controls, for having its own space.

devemux86:
I'm closing this, scale bar can be tracked in its own issue.

PR #697, opened 2015-09-20
Fix bottom margin of the attribution in the samples.

schildbach:
It introduces two dimensions to configure the appearance of the attribution more easily. The actual values are unchanged.


schildbach:
Amends #696.

devemux86:
Thanks.

PR #698, opened 2015-09-21
Set minimum Android SDK to API level 10 (Android 2.3.3).

schildbach:
API level 9 (Android 2.3-2.3.2) is not in use any more. Also, it's not compatible to Java5, forcing us to use Java 1.4 only!


devemux86:
Thanks, merged 20662b87aae55d1fdba163d6d3b455ce1291b4a0.

PR #711, opened 2015-10-24
bugfix: Way equals comparison typo

R-Amchip:
typo in equal comparison causes 2 equal but not same objects to never compare true


devemux86:
Thanks, valid report.
But can you post the PR on [dev](https://github.com/mapsforge/mapsforge/blob/master/docs/Getting-Started-Developers.md#how-to-contribute) branch?

That's our workbench, we synchronize master on special cases (like releases).

PR #712, opened 2015-10-24
bugfix: Way equals comparison typo

R-Amchip:
typo in equal comparison causes 2 equal but not same objects to never compare true
See #711


devemux86:
Thanks for the fix.

PR #715, opened 2015-10-31
Update Android buildToolsVersion to 23.0.2.

schildbach:
I propose to consider this for backporting to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ jdk:
   - oraclejdk7
 android:
   components:
-    - build-tools-23.0.1
+    - tools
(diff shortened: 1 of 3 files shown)


schildbach:
Hmm, Travis seems to access an old repository. I'll investigate…

ludwigb:
It probably makes little difference, but I just upgraded the server to have
23.0.2.
If you have a solution for travis, we can merge this in.

devemux86:
See [here](https://github.com/travis-ci/travis-ci/issues/3062).

Though I prefer our Travis configuration to be clean.
I thought that tool updates are automatic or maybe they need some time to update.

Issue #645, opened 2015-07-06
Invalid Number of Way Nodes


lincomatic:
It is a common occurrence to get a warning message from this block in MapFile.java:

			if (numberOfWayNodes < 2 || numberOfWayNodes > Short.MAX_VALUE) {
				LOGGER.warning("invalid number of way nodes: " + numberOfWayNodes);
				// returning null here will actually leave the tile blank as the
				// position on the ReadBuffer will not be advanced correctly. However,
				// it will not crash the app.
				return null;
			}
with numberOfWayNodes == 0. I'm guessing it's a bug in the map writer. An example is in japan.map, in the block where z=4,x=13,y=6

devemux86:
I'm aware of this and see it often, specially on custom maps with a lot of complex lines / polygons.
I intend to check what actually happens under the hood.

ludwigb:
We should check why the map writer writes 0 length ways into the file and probably remove them. But I suggest for stability that this is something past the next release (we do not crash at the moment, even though this is undesirable).

devemux86:
+1

lincomatic:
It's very strange... there is code in TDWay.java that specifically checks for a valid number of nodes before creating a TDWay:

		// only ways with at least 2 way nodes are valid ways
		if (way.getWayNodes().size() >= 2) {
			boolean validWay = true;

I don't understand how the 0 node ways are getting written out.

ludwigb:
Are we maybe having an overflow in the reader?
Am 30.10.2015 08:00 schrieb "lincomatic" <notifications@github.com>:

devemux86:
I haven't check out yet, but I'm able to quite easily reproduce the issue on maps with complex polygons, like isopleths.

ludwigb:
Would be good to identify this.

lincomatic:
I don’t think so. Code:

 

int numberOfWayNodes = this.readBuffer.readUnsignedInt();
if (numberOfWayNodes < 2 || numberOfWayNodes > Short.MAX_VALUE) {
   LOGGER.warning("invalid number of way nodes: " + numberOfWayNodes);

ludwigb:
Emux,
can you share one of the polygons that trigger the problem?

devemux86:
@ludwigb I sent a sample.

ludwigb:
Playing around with this. First funny result: changing the names of objects
in the map (my objective was to identify the culpable geometries), results
in different elements to be reported invalid.

[...]

PR #719, opened 2015-11-14
Fix invalid number of way nodes 0

lincomatic:
https://github.com/mapsforge/mapsforge/issues/645

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2010, 2011, 2012, 2013 mapsforge.org
+ * Copyright (c) 2015 lincomatic
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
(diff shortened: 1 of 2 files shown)


ludwigb:
Would be excellent if that fixes the problem.

My one concern is possibly invalid segment counts in polygons/multipolygons/relations.

Can I suggest that we test this with a few files first to see that this does not introduce any new problem?

devemux86:
@lincomatic Thanks, I'll have a look at it later.

devemux86:
@ludwigb +1

devemux86:
I can confirm that the situation has been improved.
Though I still see (more rarely) the warnings in question.

lincomatic:
Can you tell me how to reproduce the warning? I ran it with your sample dataset and didn't see a problem. Is there a specific tile?

devemux86:
I updated my answer. :smile: 
I'm playing with a multiple-map setup and forgot to update all maps.

devemux86:
I'm building regular country maps too, and not seen any new problem introduced.

Will continue the tests, if all go well we can think when we should merge the fix, since it's about the writer only.

ludwigb:
I have merged it in. If there are no concerns, we should move it to master soon and also get it into 0.6

devemux86:
@ludwigb +1

PR #796, opened 2016-03-17
Dev

ksaihtam:


devemux86:
Thanks for the PR. Some notes:

- Provide a title and/or description for people to understand what it handles inside.

- Each *independent* change should be contained in its own PR, for easy testing and acceptance.
The two commits handle different things (a fix and a feature enhancement).

- Any change should be implemented in all available code.
i.e. besides `AndroidTileBitmap`, there is also `AwtTileBitmap` with same interface.

- These can be done easily via creating a branch for each new PR feature or bug fix.

ksaihtam:
> Any change should be implemented in all available code.

In master branch this is already fixed in AndroidTileBitmap but not in AwtTileBitmap

devemux86:
I think we can close it now.

Issue #795, opened 2016-03-16
Polygon Corruption


lincomatic:
Cemetary polygons in the Berlin map below are corrupted. I have traced back using old map writers, and v0.4.0 and earlier render them correctly, while v0.5.1 and newer cause the error. Some change to the polygon processing code of the map writer in between those versions created the bug.

![47a1f5b2-d59d-11e5-8ca2-12036177ea89](https://cloud.githubusercontent.com/assets/1422186/13832092/f3ca673a-eb95-11e5-9c22-789bde32b49c.jpg)

applantation:
Can you identify the OSM object that does cause this? We have to see what
is strange about it.

lincomatic:
http://www.openstreetmap.org/way/5096178
and
http://www.openstreetmap.org/way/137041447

lincomatic:
Narrowed it down .. 0.5.0 is OK, so the bug creeped in between 0.5.0 and 0.5.1. How do I figure out which changes happened in this range?

lincomatic:
I reverted the JTSUtils.java portion of this commit, and the problem goes away:

https://github.com/mapsforge/mapsforge/commit/7bda373c862b8b8c8b013a12ae33cc36bfe36ef6

lincomatic:
While reverting the commit seems to fix the problem, the underlying issue seems to occur elsewhere, possibly in the code that cuts the polygons. I will research this some more and report back.

devemux86:
Probably related to #788.

lincomatic:
OK, I found the cause, now we need to find a good solution. The commit [7bda373](https://github.com/mapsforge/mapsforge/commit/7bda373c862b8b8c8b013a12ae33cc36bfe36ef6) relies on the shape of TDWay to be accurately determined in this line:

`if (way.getShape() != TDWay.LINE || innerWays != null && innerWays.size() > 0) {`

The problem is that the underlying code doesn't properly categorize TDWay, because OSMUtils.isArea() doesn't properly flag cemeteries as polygons. I added this code to isArea():

```
                if ("landuse".equals(key)) {
                    return true;
                }
```

and the error goes away. So the underlying problem goes back to what the commit was trying to fix. We don't have a foolproof way of determining whether or not a way is a polygon or not. isArea() is hard to fix .. adding my landuse clause is only a band aid, because there probably other tags that we missed.

[...]

Issue #788, opened 2016-02-24
Mapwriter 0.6 missing areas


Klaus-Tockloth:
With mapwriter 0.6 I have the problem that sometime some areas are missing. It's not a general problem because mostly the areas are rendered. The problem doesn't occur with mapwriter 0.5. 

With mapwriter 0.6 (landuse=cemetery and leisure=sports_centre are missing):
![bildschirmfoto 2016-02-24 um 20 51 33](https://cloud.githubusercontent.com/assets/5493128/13298786/7bfe1944-db38-11e5-96dc-c2b4825205d4.jpg)

With mapwriter 0.5:
![bildschirmfoto 2016-02-24 um 20 50 44](https://cloud.githubusercontent.com/assets/5493128/13298795/875de0e4-db38-11e5-9cdf-ead8f6e2178d.jpg)

Link to OSM: https://www.openstreetmap.org/#map=16/51.9580/7.6075

Regards Klaus

devemux86:
Simply the map-writer jar changes between the 2 procedures, nothing else?

Klaus-Tockloth:
Yes, I only exchanged the mapwriter jar.

applantation:
two possible culprits here:

https://github.com/mapsforge/mapsforge/commit/d10d3fab82ff6f1d61308efda85b589ddb2283c8

https://github.com/mapsforge/mapsforge/commit/c9ac9df22fdaace73a43c39f237721dbc8761491

I would not think that the multi-language features have anything to do with
it.

Anything special about that cemetery area in question?

Ludwig

Klaus-Tockloth:
> Anything special about that cemetery area in question?

https://www.openstreetmap.org/way/11784438
barrier=fence
landuse=cemetery
name=Zentralfriedhof
website=http://www.zentralfriedhof-muenster.de/
wikipedia=de:Zentralfriedhof Münster

https://www.openstreetmap.org/way/40924122
barrier=fence
leisure=sports_centre
name=Sportpark Sentruper Höhe

https://www.openstreetmap.org/way/157045098
barrier=fence
landuse=cemetery
name=Friedhof Steglitz
wikipedia	de:Friedhof Steglitz

My first idea was, that the tag "barrier=fence" has something to do with the issue.
But on the hand side there are areas with the same tagging schema which are drawn correctly:

https://www.openstreetmap.org/way/151353918
barrier=fence
landuse=cemetery
name=Friedhof Eythstraße
[...]

PR #797, opened 2016-03-17
Fix for #795 and #788?

lincomatic:


devemux86:
Related to #795 and #788.

PR #798, opened 2016-03-20
Fix wrong tile bitmap expiration check.

ksaihtam:
This patch fix the wrong expiration check in the tile bitmap objects.


devemux86:
Thanks!

PR #799, opened 2016-03-20
Fix a race condition in Layer-Manager.

ksaihtam:
This patch fix a race condition in the layer manager if layers are added/removed and the worker walk through the layer list.


devemux86:
Thanks!

I created #800 as we have seen race conditions during map state changes.
We'll monitor there if there are going to be more reports.

PR #801, opened 2016-03-20
Fix race condition on start/stop. Change cancel executors.

ksaihtam:
This patch fix a race condition on start/stop in the mapworker pool. It will stop the executors in two phases like in the oracle doc [link](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html). The changes in JobQueue will speedup the stop process for 200ms in the worst case.

With this i can create and recreate my layers without any memleak forced by the mapworkerpool and his executors. I will monitor this but i think this will fix it.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/MapWorkerPool.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/MapWorkerPool.java
@@ -21,15 +21,18 @@ import org.mapsforge.map.layer.Layer;
 import org.mapsforge.map.layer.cache.TileCache;
 import org.mapsforge.map.layer.queue.JobQueue;
 
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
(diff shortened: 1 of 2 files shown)


devemux86:
Thanks, will check it.

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/MapWorkerPool.java
```diff
@@ -61,18 +65,52 @@ public MapWorkerPool(TileCache tileCache, JobQueue<RendererJob> jobQueue, Databa
...
+                }
+            }
+        } catch (InterruptedException e) {
+            LOGGER.log(Level.SEVERE, this.LOGTAG + "shutdown self executor interrupted", e);
+        }
+
+        try {
+            if (!this.workers.awaitTermination(100, TimeUnit.MILLISECONDS)) {
+                this.workers.shutdownNow();
+                if (!this.self.awaitTermination(100, TimeUnit.MILLISECONDS)) {
```
devemux86: Here it should probably be `this.workers`.

devemux86:
Merged via a34fed09296666d0cdfcaedf8fe00302e9597050 with minor improvements.
(related to #804)

PR #809, opened 2016-03-25

mjcross:
Add explanatory bit masks for Flag bits in file header, to make it clearer how to interpret the UInt8 value


devemux86:
Thanks!

mjcross:
You're welcome - I'm 90% of the way through writing a parser for the file headers in Swift 2.0. Parsing the varint's in a little endian environment has been 'educational' LOL. Will share the code when I'm happy with it :)

devemux86:
Is this part of a larger project for Mapsforge on iOS ?

mjcross:
Yes indeed - I invested a major effort a couple of years ago, to convert and style all the open source vector mapping of the UK released by the Ordnance Survey into OSM XML, so that I could use mkgmap to get it working on a Garmin GPS (freely available via [http://bit.ly/1TdqtWi]).Then a contact pointed out that with only a little more work I could encode it to mapsforge for Android users (cue a lot more time working on the styling). I have almost 500 users for that now - but it has been bugging me that I can't run it natively on my iOS devices. It's possible (just) via the 'vectorialmap' app, but I find that a bit sluggish.
SOoooo... got myself set up with a mac mini and Xcode in the autumn and have been steadily teaching myself Swift and iOS app development. Flirted briefly with the approach of using J2ObjC on the library, but don't need the full API so (probably unwisely) I've decided to have a crack at coding it natively instead to see how far I can get... so far so good, but a long way to go. One step at a time gets you a long way eventually though :)

devemux86:
Thanks for the feedback, keep us informed about the progress!

Also in the forum you can find references about other iOS attempts.
I has posted recently a [question](https://groups.google.com/forum/#!topic/mapsforge-dev/A1CpRxcwQtI) about current state of such projects.

PR #810, opened 2016-03-26

mjcross:
Add comments to (1) clarify interpretation of the five-byte value in the Tile Index Entry; and (2) give a hint on how to use slippy map tile coords to work out how many tile index entries there are in a subfile.


mjcross:
Whoops!  Just noticed I've made an error in that... sorry: the mask for the 'sea tile' bit should actually be 0x80 00 00 00 00. Stupid boy!

PR #811, opened 2016-03-26

mjcross:
Add comments to (1) clarify the interpretation of the 5-byte value in the  tile index entry; and (2) give a tip on how to work out how many index entries there will be, by converting the bounding box to slippy map tile coords.

### DIFF
--- a/docs/Specification-Binary-Map-File.md
+++ b/docs/Specification-Binary-Map-File.md
@@ -91,7 +91,7 @@ To read the data of a specific tile in the sub-file, the position of the fixed-s
 
 |**bytes**|**optional**|**name**|**description**|
 |---------|------------|--------|---------------|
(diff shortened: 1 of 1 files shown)


devemux86:
Thanks!

BTW you can fix your PR and at the end [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits into one.

mjcross:
I can what-who-when? Heh - I'm a rank amateur at this GitHub business; it's a miracle I managed to submit anything at all (!) :D

devemux86:
Probably the simpler approach for the submitter in order to fix any issues in a pull request is to close it, and submit a new one.

You'll see in the future that another approach is appending follow-up commits to the pull request which fix issues introduced by commits earlier in it. But leaving them that way is almost always a really bad idea, because it violates the rule of one logical change per commit.
So amending / [squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits and force push in the pull request keeps the history clean.

PR #816, opened 2016-04-03

mjcross:
add links providing further details of how to encode/decode VBE-U and VBE-S formats (a.k.a. Little Endian Base 128)


devemux86:
Thanks!

PR #818, opened 2016-04-06
Added a cache initialization flag to get the current state of the int…

ksaihtam:
This flag give us the current state, mainly for the file system cache, of the initialization process. I think this is helpful for slow file systems with a initialization latency of some seconds.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/cache/FileSystemTileCache.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/cache/FileSystemTileCache.java
@@ -91,6 +91,9 @@ public class FileSystemTileCache implements TileCache {
                     }
                 }
             }
+            synchronized (initDone) {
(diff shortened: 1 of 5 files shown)


devemux86:
Thanks, some first remarks:
- I wonder if it's the best way to be notified for cache initialization with a flag instead of a listener or an observer like we use in other places in the library.
- For variable declarations and method returns you can use primitives, i.e. `boolean` and not `Boolean`.
- Also try inserting variables and methods in a sorted order (old Mapsforge habit).

ludwigb:
observer is better, otherwise user will have to poll

devemux86:
I created #819 where discussion could take place first [prior](https://github.com/mapsforge/mapsforge/blob/master/CONTRIBUTING.md) any implementation.

devemux86:
I think we can close this PR as it plays with flag setting.
Like mentioned above, a better implementation would be via observer.

PR #828, opened 2016-04-26
Fix bug handling vertical lines

lincomatic:


devemux86:
What's the case on this?

Doesn't the calculated point has to be based on the start point?

lincomatic:
It's used by the way text rendering code. I had a certain way where the name rendered vertically offset instead of over the way. 

Hitting the bug is probably very rare due to double precision floats. But in my own maps, the format is slightly modified .. stored in Mercator and with somewhat reduced precision to increase speed and make the data more compact, so it is more likely to have the bug pop out.

devemux86:
Thanks, merged via a3a500298555e2dae4327277302bea6cb66c3317.

Issue #647, opened 2015-07-10
Java documentation


devemux86:
Ongoing task of improving javadoc comments.

emdete:
addMapDataStore has two booleans and i'm a bit unsure how to use these. can you improve that doc?

PR #841, opened 2016-06-26
Document layer add/clear/remove methods

mvglasow:
Adding some Javadoc that I've recently been looking for.

Note that Eclipse inverted the sorting of the imports on save (default Mapsforge settings), hence the extra change in the commit.


devemux86:
Thanks!

Related to #647.

Issue #806, opened 2016-03-21
blank tiles with Multithreaded Rendering


calimoto-GmbH:
if multithread rendering is enabled a lot of blank tiles randomly appear:

![2016-03-21 13 53 38](https://cloud.githubusercontent.com/assets/9555809/13920771/048e143a-ef75-11e5-93c0-39f396b4c05a.png)
![2016-03-21 13 53 23](https://cloud.githubusercontent.com/assets/9555809/13920773/06e4e92a-ef75-11e5-9c33-6bd5a122a9e4.png)

and a lot of warnings in the mapfiles are logged:

invalid number of way coordinate blocks: 0
invalid way tag ID: 10496

to prove just use the samples app and increase the render Threads in the settings.

devemux86:
Multi-threading rendering is normally not recommended (default is 1 thread), as there are [issues](https://github.com/mapsforge/mapsforge/issues/591).

Did you try with the default (in Samples) persistent cache enabled?
Because there is an [issue](https://github.com/mapsforge/mapsforge/issues/764) about missing tiles with that.

SebastianDa:
yes both persistent cache (enabled and disabled) behave the same.

devemux86:
Closing in favor of #591.

bvgastel:
Actually the blank titles appears to be due to race conditions in MapFile. Multiple render threads are reading data from a ```readBuffer``` attribute in MapFile that is set differently by each thread, therefore the threads interfere with each other. I can contribute a fix for this if requested (have it working).

devemux86:
@bvgastel if you have something working all will be benefit from an PR!

bvgastel:
I have made a PR. Hope it helps! If you need more information, please let me know.

devemux86:
Also note that the black tiles usually appear with large number of threads.

bvgastel:
Black or blank?

devemux86:
Sorry blank (I corrected it).

bvgastel:
:)

It can take a while to fill in the blank tiles, but the blank tiles are eventually painted in with map data (as opposed to the old behaviour that the blank tiles are not replaced).

devemux86:
@calimoto-GmbH could you check the latest master?

devemux86:
Closing here, if there any issues please report.

DrMarkEy:
[...]

Issue #591, opened 2015-03-20
Multithreaded Rendering: rendering errors


ludwigb:
The new multi-threaded renderer does not play well with labels/icons across tile borders. 

For the original discussion see: 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mapsforge-dev/x19w0iPealM/K3SgbhwdcowJ

ludwigb:
With c16993bc7c003445 I have just pushed a fix for this. 

The problem is that the LabelCache was not synchronized properly and that also tiles that were in the process of being rendered were not taken into account for the label boundary overlap calculations. 

With a threaded TileStore there is still a window for this to happen. As with a multi-threaded renderer the TileStore does not need threading anymore, this should be removed, closing this window. With that done it will also be possible to simplify the synchronization with the LabelCache a bit.

devemux86:
With ff0ca396ea4dcc45ee67e592adc00e75fba07a70 I pushed a fix for NPE happening with ```LabelLayer```.

devemux86:
About the multi-threaded renderer so far the tests work fine (threaded cache off).

I have seen once some rendering artifacts that went away when I reloaded the map.
Somehow part of a tile showed blurry graphics, like above zoom level.

devemux86:
Playing with 8 threads this tiles mix showed up:

![8_threads](https://cloud.githubusercontent.com/assets/3484020/6765606/810afaa0-cff0-11e4-9c9f-e3138a3d3a72.png)

devemux86:
It's very easy to mix the tiles with consecutive pan / zoom actions (screenshot).
Playing with default settings of 6 threads.

Also with large maps I have very often app crashes at small zoom levels (whole country view).. 

![sample](https://cloud.githubusercontent.com/assets/3484020/6770187/a1dceb18-d0c1-11e4-885d-ca9971017f84.png)

devemux86:
And the exception from the app crash:
```java
Caught a RuntimeException from the binder stub implementation.java.lang.NullPointerException
    at android.inputmethodservice.IInputMethodWrapper.setSessionEnabled(IInputMethodWrapper.java:280)
    at com.android.internal.view.IInputMethod$Stub.onTransact(IInputMethod.java:129)
    at android.os.Binder.execTransact(Binder.java:404)
    at dalvik.system.NativeStart.run(Native Method)
```

ludwigb:
I think we are hitting two things here: 
[...]

PR #863, opened 2016-08-12
fixing race conditions

bvgastel:
These race conditions resulted in blank tiles in wrong places, and titles appearing on wrong places. Related to issue #806 and some of the issues mentioned in #591 (which actually consists of multiple separate problems).

There is another possible improvement to MapFile possible. To alleviate a memory problem, ReadBuffer can instead of reading all the relevant data only read in a window of 128 Kb or so, and move this window. This cuts down on the memory needed, but costs somewhat more system calls.

But for this PR I have minimised the changes. Hope it helps!


devemux86:
A PR should include only the minimum amount of changes for its improvement / fix.

Now we a series of commits with many changes, probably because you created the PR based on your master branch.

bvgastel:
Oh, sorry, I did not intent to include those. I did not expect that new updates are automatically included in a PR (as I'm more familiar with GitLab). I will try to fix the PR.

bvgastel:
closing this one, in favour of the one with a separate branch.

PR #864, opened 2016-08-13
fixing race conditions

bvgastel:
These race conditions resulted in blank tiles in wrong places, and titles appearing on wrong places. Related to issue #806 and some of the issues mentioned in #591 (which actually consists of multiple separate problems).

But for this PR I have minimised the changes. Hope it helps!

### DIFF
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/IndexCache.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/IndexCache.java
@@ -86,9 +86,11 @@ class IndexCache {
             int indexBlockSize = Math.min(SIZE_OF_INDEX_BLOCK, remainingIndexSize);
             indexBlock = new byte[indexBlockSize];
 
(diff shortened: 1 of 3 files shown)


devemux86:
Thanks!
Why the need of passing the `ReadBuffer` as a parameter in all methods, instead of the current situation with a central variable?

bvgastel:
No problem.
The passing is because the ReadBuffer object maintains a state that is changed by all the concurrent threads. So thread 1 sets the ReadBuffer to some value, thread 2 to some other value, while thread 1 still expects the ReadBuffer to contain the value set by thread 1. To avoid these problems, each thread must have its own ReadBuffer object. The easiest manner is to avoid a central variable and pass it along each time.

devemux86:
Being a core change I merged it for now in `issue_864` branch for some testing.

devemux86:
Merged via 5afb51e1b1fbeb90cdd62ffd22dabbf4c3e7574d (adding your name and some documentation).

PR #884, opened 2016-10-14
Polygon label position calculation enhancements

andreynovikov:
1. Implements per-tag configuration for label position calculation.
2. Enhances label position calculation by utilizing the following algorithm:
https://www.mapbox.com/blog/polygon-center/

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/PolyLabel.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/PolyLabel.java
@@ -1,0 +1,251 @@
+/**
+ * Copyright 2016 Andrey Novikov
+ * Java implementation and adoption for JTS and MapsForge of
+ * https://github.com/mapbox/polylabel
+ * <p>
+ * ISC License
+ * Copyright (c) 2016 Mapbox
+ * <p>
+ * Permission to use, copy, modify, and/or distribute this software for any purpose
+ * with or without fee is hereby granted, provided that the above copyright notice
+ * and this permission notice appear in all copies.
+ * <p>
+ * THE SOFTWARE IS PROVIDED "AS IS" AND ISC DISCLAIMS ALL WARRANTIES WITH REGARD TO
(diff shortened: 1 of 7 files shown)


mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/PolyLabel.java
```diff
@@ -0,0 +1,251 @@
...
+import com.vividsolutions.jts.geom.Point;
+import com.vividsolutions.jts.geom.Polygon;
+
+import java.util.Comparator;
+import java.util.PriorityQueue;
+import java.util.logging.Logger;
+
+public class PolyLabel {
+    private static final Logger LOGGER = Logger.getLogger(PolyLabel.class.getName());
+    private static final float SQRT2 = (float) Math.sqrt(2);
```
devemux86: Why casting to 'float'?

mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/PolyLabel.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final double precision = 0.00000011920928955078125; // one logical pixel for zoom 14
```
devemux86: Can you post in the comment the formula of the result?

andreynovikov: 1 / (512 << 14)

mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/PolyLabel.java
andreynovikov: I was writing the code with floats first and then switched to doubles because JTS operates with doubles. Should be `double SQRT2 = Math.sqrt(2.0)` of course.

devemux86:
Thanks!
I merged the PR in [issue_884](https://github.com/mapsforge/mapsforge/tree/issue_884) branch for testing.

devemux86:
@andreynovikov thanks!
Merged via f2090f4d52fefbb9b910bfb7b0f5e1fb1d979a47.

PR #890, opened 2016-10-25
Simplification-max-zoom; ZoomAnimation and MoveAnimation is same thread now

mikes222:
Simplification-max-zoom implemented as parameter in the map-writer. 
Zoomanimation and movement-animation works now in the same thread.

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
@@ -150,7 +150,7 @@ public final class MapFileWriter {
 
             // TODO is this the right place to simplify, or is it better before clipping?
             if (this.configuration.getSimplification() > 0
-                    && this.tile.getZoomlevel() <= Constants.MAX_SIMPLIFICATION_BASE_ZOOM) {
+                    && this.tile.getZoomlevel() <= this.configuration.getSimplificationMaxZoom()) {
                 processedGeometry = GeoUtils.simplifyGeometry(this.way, processedGeometry, this.maxZoomInterval,
                         tileSize, this.configuration.getSimplification());
                 if (processedGeometry == null) {
@@ -908,7 +908,7 @@ public final class MapFileWriter {
                     if (processedTiles == amountTiles) {
                         LOGGER.info("written 100% of sub file for zoom interval index " + zoomIntervalIndex);
                     } else {
-                        LOGGER.info("written " + (processedTiles / amountOfTilesInPercentStep) * PROGRESS_PERCENT_STEP
+                        LOGGER.info("written " + Math.round(processedTiles / amountOfTilesInPercentStep / PROGRESS_PERCENT_STEP * 100)
                                 + "% of sub file for zoom interval index " + zoomIntervalIndex);
                     }
                 }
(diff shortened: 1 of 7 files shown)


devemux86:
>Simplification-max-zoom implemented as parameter in the map-writer.

Useful option. :+1: 

>Zoomanimation and movement-animation works now in the same thread.

Will check this!

BTW it's [recommended](https://help.github.com/articles/about-pull-requests/) using a topic branch for each pull request.
Having that, each feature should be handled in its own pull request for easier merging.

mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
```diff
@@ -250,16 +264,6 @@ public synchronized void init(PreferencesFacade preferencesFacade) {
...
-     * Animates the center position of the map by the given amount of pixels.
-     *
-     * @param moveHorizontal the amount of pixels to move this MapViewPosition horizontally.
-     * @param moveVertical   the amount of pixels to move this MapViewPosition vertically.
-     */
-    public void moveCenter(double moveHorizontal, double moveVertical) {
-        this.moveCenterAndZoom(moveHorizontal, moveVertical, (byte) 0, true);
-    }
-
-    /**
```
devemux86: Why different implementation specifically for `moveCenter(moveHorizontal, moveVertical)` ?

mikes222:
Because the original implementation moves the position (with no animation) and starts the zoomanimator with to do (zoomlevel is the same as before). 
The former is a desired behavior the latter not. 

As I recheck the code I would also recommend to change the implementation in 
> public void moveCenter(double moveHorizontal, double moveVertical, boolean animated)

and

>public void setMapPosition(MapPosition mapPosition, boolean animated)

accordingly.

If the animation flag is meant to animate the movement the methods should call animateTo() instead of setCenterInternal().

devemux86:
I think that `animated` parameter was introduced only for zoom.
Nevertheless the internal API could use some improvements.

mikes222:
You are right. It would be better to use the current behavior for existing methods and maybe introduce another family of "animateTo" methods for convenience.

mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
```diff
@@ -908,14 +908,15 @@ private static long writeSubfile(final long startPositionSubfile, final int zoom
...
-                        LOGGER.info("written " + (processedTiles / amountOfTilesInPercentStep) * PROGRESS_PERCENT_STEP
+                        LOGGER.info("written " + Math.round(processedTiles / amountOfTilesInPercentStep / PROGRESS_PERCENT_STEP * 100)
                                 + "% of sub file for zoom interval index " + zoomIntervalIndex);
                     }
                 }
 
                 // TODO accounting for progress information
             } // end for loop over tile columns
         } // /end for loop over tile rows
+        LOGGER.info("written 100% of sub file for zoom interval index " + zoomIntervalIndex);
```
devemux86: We shouldn't need that, as is causes double logs for 100%.

devemux86:
Thanks!
Merged partially via 70cd87ac7dd31f1a29fc1bdd55bfefb1ab84e8e7 and map writer via dcfa789ceaaa7ce351ec464c5078698523871907.

devemux86:
Merged also `MapViewPosition` via 036e82276d4a9ba00751318da26a4879c6c5badf.
Not included the `moveCenter` change, better seeing that in #893.

PR #899, opened 2016-11-16
Added hint on search by tag, added link to POI databases

wish7code:
So far the documentation didn't mention the POI search by tag option, I added small hint, which might be useful..

Cheers
Toby


devemux86:
Thanks!

Issue #889, opened 2016-10-23
Customize mapView animation speed and interpolation


bruneric:
Sometimes i use animateTo() method to animate the mapView to a specified position, and it works very well. Unfortunately i am not able to customize the animation speed(or duration), and i can't find a way to change the default interpolation of the mapView animation. My question is: is it possible to customize speed and/or interpolation of mapView's animateTo() method?

devemux86:
`MapViewPosition.animateTo` uses some predefined constants for the animation calculation.
We could add another method with these as variables or even come with a better implementation. :slightly_smiling_face:

PR #902, opened 2016-11-23
MapViewPosition animator refactored

mikes222:
Animation refactored. Now it is done by creating animation-jobs which will be queued and executed one after another. In my tests this is much faster and more stable than the original implementation.
Cool thing is if the job is delayed for some reason it just processes a few steps or even just the final animation step.
Furthermore I made the solution for issue #889 too.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
@@ -70,11 +70,6 @@ public class MapViewPosition extends Observable implements Persistable {
             }
         }
 
-        /**
-         * This method is not totally correct since we also should check if a job is currently being processed.
-         *
-         * @return
-         */
         public boolean hasWork() {
             return working.get() || jobs.size() > 0;
         }
@@ -553,9 +548,9 @@ public class MapViewPosition extends Observable implements Persistable {
             } else {
                 setCenterInternal(newLatitude, newLongitude);
                 setZoomLevelInternal(this.zoomLevel + zoomLevelDiff, animated);
+                notifyObservers();
             }
         }
-        notifyObservers();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


devemux86:
Thanks, will review!

devemux86:
Have not checked the code, but playing with map gestures I see several issues.

- At zoom in (swipe or double tap) the animation ends jerkily by shaking the map before the final rendering
- Slightly swipe for zoom in / out and the map stays in between instead of completing the gesture towards the next / previous zoom level (and if we slightly zoom in / out again we're getting in a loop)
- All these need extremely thorough testing for not affecting negatively users experience :slightly_smiling_face: 

Can you explain in detail what's the purpose of the changes, fixing existing issue (what?) or add new functionality?

Since it alters how users are accustomed of map mechanics (e.g. new action stops previous), it's preferable to not break old API implementation and add any new functions as optional.

Even better I'm thinking if it's more appropriate to introduce a 2nd `MapViewPosition` if we're going for a different behavior.

mikes222:
We could introduce the new implementation as new MapViewPositiion and declare it as "experimental". If you want me to do so I will. 
But besides that I need to fix the issues in the new implementation anyway. 

I have experienced some flaws with the current implementation especially at high cpu-usage when move/zoom was triggered by the program: 
- Sometimes the map zoomed in/out in a loop
- Not all tiles had been drawn sometimes. 
- Zoom or move had not been performed sometimes. 
- Flickering when moving via animation

With the new implementation mapsforge does what it should do when the program requests it. However according to your tests it is not always doing what the user wants. I can confirm your issues and I will fix them in the next days. 

Is there any possiblity to add the changes to the pull request or do I need to close this pull request and create another one?

devemux86:
Besides higher cpu at consecutive map position changes via code, I have not seen anything else so far with existing implementation.
The preferred way is always to report first any problems with detailed instructions for how we can reproduce them and then see how to proceed with fixes.

I see you created the pull request from your master branch, that is not [recommended](https://help.github.com/articles/about-pull-requests/) for various reasons. So better creating PR from a topic branch, which you can rebase on latest master in the meantime.

Regarding the implementation it's preferable leaving existing class, as current library behavior is workable already by users. We don't want to break their experience (except bug fixes).

So how about extending `MapViewPosition` in a new class where you can implement the new logic and then see how it goes?

devemux86:
Thanks for the work!
Closing here and can discuss any improvements like mentioned above in a new PR.

PR #943, opened 2017-03-27
Pin and PinGroup

virus-warnning:
Add two classes Pin and PinGroup.

* Pin is a layer like Marker, but it doesn't need bitmap. 
* PinGroup is a container of Pin, and let all Pins work in single selection mode.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/overlay/Pin.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/overlay/Pin.java
@@ -1,0 +1,163 @@
+package org.mapsforge.map.layer.overlay;
+
+import org.mapsforge.core.graphics.Canvas;
+import org.mapsforge.core.graphics.Color;
+import org.mapsforge.core.graphics.FontFamily;
+import org.mapsforge.core.graphics.FontStyle;
+import org.mapsforge.core.graphics.GraphicFactory;
+import org.mapsforge.core.graphics.Paint;
+import org.mapsforge.core.graphics.Path;
+import org.mapsforge.core.model.BoundingBox;
+import org.mapsforge.core.model.LatLong;
+import org.mapsforge.core.model.Point;
+import org.mapsforge.core.util.MercatorProjection;
+import org.mapsforge.map.layer.Layer;
+
+/**
+ * Pin of POI
+ * 
+ * This Layer draw a pin and some text on the map directly.
+ * The pin has selected and unselected state, using two different colors to represent.
+ * Under selected state, a label string appears below the pin to explain what it is. 
+ * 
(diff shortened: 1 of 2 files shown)


virus-warnning:
pin-and-pingroup

devemux86:
Thanks!
Some first notes:

- We better avoid creating paints in each `draw` call and instead pass them as parameters for performance.
- All fixed values (like sizes) in calculations better use the display model scale factor for playing well in all device densities.

e.g. you could check the [debug layers](https://github.com/mapsforge/mapsforge/tree/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/debug) for how we handle such cases.

Also could you add an example in [mapsforge-samples-android](https://github.com/mapsforge/mapsforge/tree/master/mapsforge-samples-android) to demonstrate the use of the classes?

virus-warnning:
GraphicFactory.createPaint() seems return existed Paint.
When I create Paint to draw Pin, than Grid is also affected.

devemux86:
> GraphicFactory.createPaint() seems to return existed Paint.

What existed? A new class is created with every `createPaint` call.
You can check the other layers in [overlay](https://github.com/mapsforge/mapsforge/tree/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/overlay) package for how the paints are used.

> Is demo activity necessary in this pull request?

Well before merging something we have to test it first. :slightly_smiling_face: 
Also users will benefit from examples to see in action how new API works.
I can probably create the example later.

virus-warnning:
Thanks, i'll review more to understand how overlay works.

virus-warnning:
Is it possible to let paint support matrix, so that I can rotate pin against the azimuth.
Canvas just support drawBitmap() with matrix.

devemux86:
What do you mean?
Paint just holds style and color information about drawing, which is performed by Canvas.

virus-warnning:
Canvas.drawPath() & Canvas.drawCircle() cannot work with Matrix.

devemux86:
Canvas calls are implemented per platform in `AndroidCanvas` and `AwtCanvas` using there each native platform API.
So besides `drawBitmap` are there any other platform draw methods that can use a `Matrix` (e.g. in [android.graphics.Canvas](https://developer.android.com/reference/android/graphics/Canvas.html))?

virus-warnning:
https://developer.android.com/reference/android/graphics/Canvas.html#setMatrix(android.graphics.Matrix)

devemux86:
That's something different and Android specific as it affects the whole Canvas instead of just the draw call.
If there is need we could see it separately in another issue as we need to implement it somehow for Desktop too.

devemux86:
Closing due to lack of follow up.
Can post new PR with improvements.

PR #945, opened 2017-04-03
Add highways to POI search; Add tool geo-tagger

Gustl22:
Provide highways, to potentially extend to simple address search. The main reason to leaving out highways was the size, but it's marginal noticeable. Further you can exclude them via filters.

Added a tool to automatically tag POIs with "is_in" values (if doesn't already exists), to link POI with its region. 
Usually there's no simple way to get info about location, after conversion to POI-database. (Now, only administration level 9 is considered, furthers will follow).


devemux86:
Thanks!

BTW it would help us in reviewing if any features would be posted as separate pull requests (e.g. highways and `is_in`).
Also it'd be convenient for you to use a topic branch for the pull request (see [GitHub help](https://help.github.com/articles/about-pull-requests/)).

Gustl22:
Ok i'll fix that ;D Thx.

PR #946, opened 2017-04-03
POI: add geo tags

Gustl22:
Added a tool, which automatically tag POIs with `is_in` values (if it doesn't already exists), to link POI with its region.
Usually there's no simple way to get info about location, after conversion to POI-database (no geo-coding info available). This will solve the problem. Administration boundary levels 7 to 10 are considered. Furthers can easily attached, but they will cause much overhead for POIs, which are already tagged by higher levels. Tool is may extendible to address tags.
To turn tool on add `geo-tags=true` in command line.

### DIFF
--- a/mapsforge-poi-writer/build.gradle
+++ b/mapsforge-poi-writer/build.gradle
@@ -2,6 +2,7 @@ dependencies {
     compile project(":mapsforge-poi")
     compile 'com.vividsolutions:jts-core:1.14.0'
     compile 'org.xerial:sqlite-jdbc:3.16.1'
+    compile 'com.google.guava:guava:20.0'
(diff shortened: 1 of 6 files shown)


devemux86:
Thanks!

Merged via d8b087991f6dcd0a318659a157408d696314546c.

devemux86:
@Gustl22 can you explain more the mechanism?
e.g. which POI obtain the `is_in` tag, because there are many which still don't have it.

Gustl22:
Mechanism: It loops all relations on administration levels (10 to 7) and create a polygon of its ways. Then all POIs inside this Polygon get the is_in tag of the included place, if existent. There may not set a place or the relation has wrong ways, then it will abort. Or may i have to do a fix.

PR #947, opened 2017-04-03
POI: add non-closed ways

Gustl22:
Provide highways, to potentially extend to simple address search. An extra search environment only for addresses seems needless, cause of redundant informations and equality of both systems. And results can easily filtered by categories. _Otherwise there is need of a second map for address search with more size and this is not easy to combine with third party)_

There's no extra command to turn highways on and off, but the `ways=false` command will exclude all ways. But it would be easy to add an additional command parameter. Further you can exclude them via category filters. The filesize increase is acceptable (1/6 on top).


devemux86:
Thanks!

Merged via 286e028beeda380eb707fb568e6dd4ec52997d52 with minor improvements via b75fe1e51aee6b35bd3d02d0fe9cfabe77090f04.

PR #953, opened 2017-04-20
POI: Multiple patterns in filtered search

Gustl22:
Hello,
if you filter the poi-database by a pattern you can only use a specific one: e.g. "%name=carl%addr:postcode=111111%";
But the order of osm-tags is not guaranteed, and you'll have to create a second query to get all results. 

So i added a method with multiple patterns in AndroidPoiPersistenceManager, to support multiple subqueries on poi-data at the same time.

Sorry for the mass of commits, next time i'll use cherry pick command ;D

Sincerely,
Rebo

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.1'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
     }
 }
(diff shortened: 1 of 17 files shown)


Gustl22:
I hope u didn't start merging; found a bug, if pattern contains null element. Fixed it in last commit. Sorry for the imperfect pull-request.

devemux86:
Thanks, I'll review it soon. :slightly_smiling_face:

devemux86:
Could you create a new pull request based on latest master with just the updated classes in their own commit?

Gustl22:
Yes of course ;)

PR #958, opened 2017-05-18
Update Getting-Started-Map-Writer.md

Arman92:
Fixed incorrect argument label of preferred-language


devemux86:
Actually the option has changed from [0.6.0](https://github.com/mapsforge/mapsforge/blob/master/docs/Changelog.md#version-060-2015-11-25) to `preferred-languages`, as now map-writer supports creation of multilingual maps.

Arman92:
Oh, I didn't know, I was using 0.5.1

Issue #964, opened 2017-06-09
Android 8.0 (Oreo) ZoomButton deprecation


devemux86:
[ZoomButton](https://developer.android.com/reference/android/widget/ZoomButton.html) was deprecated in API level O.
Use other means to handle this functionality. This widget is merely a simple wrapper around a long-press handler.

devemux86:
After deprecation could copy `ZoomButton` (Apache 2.0) for user convenience.
Though if Android removes completely `ZoomControls` layout, then to provide internal zoom buttons would require image resources, i.e. turning `mapforge-map-android` to an `aar` which is outside library's scope. Like was with not including location renderer resources, for users to select their own.
So most versatile solution can be users to implement their zoom buttons, as demonstrated at rotate map sample in `mapsforge-samples-android`.

PR #969, opened 2017-06-25
ZoomButton deprecated in Android 8.0 (O) 

theaaron123:
Issue [#964](https://github.com/mapsforge/mapsforge/issues/964) changed to handle functionality without ZoomButton - delay realised using Handler postDelayed within the onClick method.

### DIFF
--- a/mapsforge-map-android/res/layout/zoom_controls.xml
+++ b/mapsforge-map-android/res/layout/zoom_controls.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+        <!--
+        /*
+        **
+        ** Copyright 2008, The Android Open Source Project
+        **
+        ** Licensed under the Apache License, Version 2.0 (the "License");
+        ** you may not use this file except in compliance with the License.
+        ** You may obtain a copy of the License at
+        **
+        **     http://www.apache.org/licenses/LICENSE-2.0
+        **
+        ** Unless required by applicable law or agreed to in writing, software
+        ** distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


devemux86:
Thanks!

Changing the class cast does not mean that the functionality will still exist.
A better solution would be copy `ZoomButton` and `ZoomControls` (Apache 2.0) to provide own independent functionality.

theaaron123:
@devemux86  sorry specifically what functionality won't work? As zoom in/out and the click delay seem to work as before, looking at the implementation of the ZoomButton class it also uses postDelayed for controlling zoom delay.

devemux86:
On Android 8.0 (O) the `ZoomButton` class exists in API and works.
In next Android any deprecated class could be removed, so [ZoomControls](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/ZoomControls.java) now having `ZoomButton` in its [layout](https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/layout/zoom_controls.xml) will change. Then will have to see the implementation.
So right now casting existing `ZoomButton` to `ImageButton` what benefit does it have?

devemux86:
See my comment in #964.

theaaron123:
@devemux86  Apologies, would you suggest to remove the ZoomButton instances from MapZoomControls and all related methods then add an example to SamplesBaseActivity.createMapViews() to create the buttons as in RotateMapViewer?

devemux86:
No need to hurry, Android 8.0 still has zoom buttons, so when they remove the functionality in next Android(s), then Mapsforge can adapt to it. :slightly_smiling_face: 

Anyway when it's time, better follow the location overlay way, where can show in Samples how users can include their own zoom buttons.

PR #970, opened 2017-06-29
An awt sample to render & save a tile.

virus-warnning:
Just draw a tile without any UI.


devemux86:
Thanks!
Merged via 5b2a868ced5ad21b324c6473a4bba15c55fd8bba.

_Please use for indentation 4 spaces (no tabs), as library [conventions](https://github.com/mapsforge/mapsforge/blob/master/.github/CONTRIBUTING.md)._

devemux86:
Regarding the `args`, they can be used for all parameters instead of fixed values.

PR #979, opened 2017-07-26
Fix and improve FrameBufferHA2

### LINKED ISSUE #977: Improved FrameBufferHA2
Improved map frame buffer.

It's active in Samples and can be enabled via:
```java
Parameters.FRAME_BUFFER_HA2 = true;
```

bailuk:
Let's give it another try!

Changes:
- Swap not only bitmaps but also bitmap requests (fixes #977)
- Move <code>allowSwap</code> from <code>FrameBufferBitmap</code> to <code>FrameBufferHA2</code>
- Reset background of newly created bitmap to background color. This fixes an issue where Android map view did display a recycled bitmap.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/controller/FrameBufferController.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/controller/FrameBufferController.java
@@ -32,7 +32,7 @@ public final class FrameBufferController implements Observer {
     /**
      * If true the {@link FrameBufferHA2} will be used instead of default {@link FrameBufferHA}.
      */
-    public static boolean FRAME_BUFFER_HA2 = false;
+    public static boolean FRAME_BUFFER_HA2 = true;
 
     private static float maxAspectRatio = 2;
 
--- a/mapsforge-map/src/main/java/org/mapsforge/map/controller/FrameBufferController.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/controller/FrameBufferController.java
@@ -32,7 +32,7 @@ public final class FrameBufferController implements Observer {
     /**
      * If true the {@link FrameBufferHA2} will be used instead of default {@link FrameBufferHA}.
      */
-    public static boolean FRAME_BUFFER_HA2 = true;
+    public static boolean FRAME_BUFFER_HA2 = false;
 
     private static float maxAspectRatio = 2;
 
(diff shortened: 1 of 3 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/controller/FrameBufferController.java
```diff
@@ -32,7 +32,7 @@
     /**
      * If true the {@link FrameBufferHA2} will be used instead of default {@link FrameBufferHA}.
      */
-    public static boolean FRAME_BUFFER_HA2 = false;
+    public static boolean FRAME_BUFFER_HA2 = true;
```
devemux86: Let's keep `FrameBufferHA2` disabled by default for now to not break user apps.

mapsforge-map/src/main/java/org/mapsforge/map/view/FrameBufferHA2.java
```diff
@@ -20,92 +20,102 @@
...
     private MapPosition lmMapPosition;
 
+    private final FrameBufferBitmap.Lock allowSwap = new FrameBufferBitmap.Lock();
+
+    private Dimension dimension;
+    private final Matrix matrix;
+
+    private final DisplayModel displayModel;
+    private final FrameBufferModel frameBufferModel;
+    private final GraphicFactory graphicFactory;
```
devemux86: Why not using `FrameBuffer` inherited variables and methods to avoid duplication?

bailuk: I see <code>FrameBufferHA2</code> as a  replacement for <code>FrameBuffer</code> and not really as a sub class. It just would make the transition easier and it's more readable.

devemux86: Such core classes need extremely good testing before become default.
Until then we can avoid copy / paste and override only the changes.

_Also please try following library code / style conventions, like sorted methods, modify methods in their existing places, etc._

bailuk:
- <code>FrameBufferHA2</code> is disabled by default
- Use inherited _variables_ and methods
- Methods are sorted

devemux86:
Thanks, seems working much better now!
Merged via 70036d703e831f198ff75ed3389add9ca6d030ca.

PR #989, opened 2017-08-31
POI: insert method isClosed:boolean into PoiPersistenceManager

Gustl22:
Check if PoiPersistenceManager is already closed
Improvements:
- Close Statements after sql(ite)-Exception
- Close ResultSets after SQL-Exception
- Check if closed, before closing it twice (android/awt)
- added synchronized to close() to avoid overall access on statements while closing


devemux86:
Thanks!
Merged via 854ed41406fef6870cd511fb3c97c4581936e5b1.

Issue #950, opened 2017-04-08
POI: many-to-many relationship among Categories and POIs


devemux86:
Implement many-to-many relationship among Categories and POIs to properly fix #940.

devemux86:
That requires a new POI specification v2.

POI specification v1 is published, used by apps and offered by e.g. [Mapsforge](http://download.mapsforge.org/) and [OpenAndroMaps](http://www.openandromaps.org/en/).

So need to support both `.poi` file versions for not breaking suddenly any backwards compatibility.
POI writer can advance totally to v2, anyone can use older jar for building v1 `.poi` - which will be deprecated by time.

But POI readers better to be able to work with both v1 and v2 versions for now.
In `metadata` table we store already the specification version, can read it at `.poi` open and follow the separate path in code.

Also let's have distinct DB statements / methods / ... with `v1` and `v2` suffixes for easy reference.

Gustl22:
In my upcoming commit #991 there's an imperceptive difference. The concerned methods have a third parameter with `version` number to avoid extra methods for each version. Often only a few lines differ from the other version, so there's no need of nearly duplicate them.

While the underlying concept is different (in new v1 and v2), but does not take effect on the queried results, and is more modular, like the following:
The DB is fragmented in:
- `metadata`, which has an exceptional position
- poi-location (`poi_index`)
- poi-tags (`poi_data`)
- `poi_categories`
- M to N relationship between POI and categories in v2 (`poi_cmap`)

Depending on the requests in `findInRect`, more tables are joined:
v1: join `poi_data`, if necessary for categories and tags
v2: join `poi_data` for tags, and `poi_cmap` for categories

Second step is the creation of POIs from the results. Only for the specific results the tags and categories are requested afterwards, to complete missing categories (v2) and tags.

The `metadata`-table must be prepared and requested before all the other statements to change them based on the poi-version. For that, this is handled in `updatePoiFileInfo`, that you cannot forget initializing its statement. This is an abstract method in `AbstractPoiPersistenceManager`, too, so `getPoiFileInfo()` could be implemented and used in that Manager.

I'm happy for some feedback... thx ;D

devemux86:
Closed via #991.

PR #991, opened 2017-09-02
POI: m to n relationship among Categories and POIs

Gustl22:
Refers to #950
- split db-queries to location, tag-data and categories 
- improve getSQLSelectString, to dynamically join data 
- PointOfInterest supports multiple categories and minor improvements (Set instead of List)
- adapt PoiPersistenceManager (multiple categories) 
- adapt PoiWriter (multiple categories)

If this pull request has too much changes in one step, i try to split them. But it doesn't make sense to separate the minor changes (as preparation for the m:n relation of categories) from the grand pull-request.
If there are any questions, i'll try to explain my thoughts / or improve my code.

### DIFF
--- a/mapsforge-poi-android/src/main/java/org/mapsforge/poi/android/storage/AndroidPoiPersistenceManager.java
+++ b/mapsforge-poi-android/src/main/java/org/mapsforge/poi/android/storage/AndroidPoiPersistenceManager.java
@@ -24,7 +24,6 @@ import org.mapsforge.poi.storage.AbstractPoiPersistenceManager;
 import org.mapsforge.poi.storage.DbConstants;
 import org.mapsforge.poi.storage.PoiCategory;
 import org.mapsforge.poi.storage.PoiCategoryFilter;
-import org.mapsforge.poi.storage.PoiFileInfo;
 import org.mapsforge.poi.storage.PoiFileInfoBuilder;
 import org.mapsforge.poi.storage.PoiPersistenceManager;
 import org.mapsforge.poi.storage.PointOfInterest;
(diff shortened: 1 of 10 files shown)


devemux86:
Thanks, will review it!

That essentially requires a new POI specification v2.
POI specification v1 is out, used by apps and offered by e.g [OpenAndroMaps](https://www.openandromaps.org/en/).
So probably could support both for not breaking backwards compatibility?
At least in readers, don't know if could be useful in writer too.

Gustl22:
Yes this definitely needs a new Version. 
If you want I can try to add a kind of version control, and depending on this, I use other query strings. Is it better to simply check the db tables or should i insert a version control with PRAGMA "user_version" ?
I don't think writer has to support the version 1, instead maybe older builds could be used to create the files. 
If you want I can handle this in a second commit ;D Sorry that i didn't inform you in the Issue-Tracker about my upcoming pull-request.

devemux86:
Agreed for map-writer and `metadata` table (that's why I first included it!).
Let's discuss it in #950, for broader visibility by users / developers.

BTW if you feel comfortable can continue work in this PR or create a fresh PR with the new concept.

Gustl22:
Now support for v1 and v2. I'm not familiar with the naming conventions of m:n relations. So feel free to rename the `poi_cmap`-table. And the `updatePoiFileInfo` may better called `load...` or `fetch...`. Don't know what's convenient.

devemux86:
Nice, will check it!

devemux86:
Nice clean work!

Since it's core change I merged it in [issue_991](https://github.com/mapsforge/mapsforge/tree/issue_991) branch with additional improvements via b1627de3875073d81ae4c2fa511e801231c62397, so we can test better.

devemux86:
I made also an [announcement](https://groups.google.com/forum/#!topic/mapsforge-dev/tmxHkE2LP4Q) in forum for larger visibility.

Gustl22:
Thanks for the improvements ;D
I opened the v2.poi database with sqlite plugin of firefox and I don't know where this table `sqlite_autoindex_poi_category_map_1` come from... I know that indexes speed up the scan through the data, but the other tables like `poi_data` and `poi_categories` don't need one. I also deleted the whole osmosis directory, to avoid temp data issues, but the table keeps existent.
image
Do you have the same result on creation? Or is this a normal process? Thx :)

devemux86:
We don't explicitly create any such index, since doing range queries via R-Tree functionality.

Probably Firefox extension cannot parse properly the table due to R-Tree nature?
Opening a v2 .poi with (recommended) [spatialite-gui](https://www.gaia-gis.it/gaia-sins/index.html) doesn't show anything strange.
Can you try with it?

Gustl22:
Yes I tried it, and the index didn't appear. Anyway slightly strange... Thx :)

devemux86:
Thanks!

Merged via 4327eb2fda82378f2d3486017baf6397d2e62344 with improvements via 2ce3237cdcb26dc41ad74113f452e442628f1d5c.

PR #995, opened 2017-09-19
POI-Writer: add bounds to metadata, if not set in parameters

Gustl22:
This is useful to check/complete the bounds of your POI file, even if no bounds were given by the parameters while POI writing. This highly depends on the converted PBF file and its inherited nodes.
First I tried to reach a similar result with `initialize(Map<String, Object> metadata)` in `MapFileWriterTask` but metadata was always empty.


devemux86:
Thanks!

devemux86:
It does not depend on input data format, as we don't use the pbf bounding box representation.

POI parsing is simpler, either use the `bbox` option or all data will be included.

Issue #923, opened 2017-02-12
Hillshading


devemux86:
Hillshading from digital elevation model data like [HGT](http://wiki.openstreetmap.org/wiki/SRTM#Format) format.

devemux86:
See PR #922 by @usrusr.

devemux86:
On Desktop Samples the 1st argument can be used as `dem` folder.

Will leave the issue open for monitoring further hillshading improvements.

demiantres:
The hillshading algorithm produces visible seams along HGT file borders:

![mfbound](https://user-images.githubusercontent.com/5980561/30467982-b7c7d968-99e8-11e7-9d90-0ca55e991968.png)

usrusr:
Visible seams in rough areas are known (interpolation artifacts: the interpolator assumes that the neighboring pixel outside the texture is the same color as the closest available, the dark part at the northern hill is one of those), But the dark line in the lowlands must be something different. 

Except if the screenshot is from a build based on the stale state of my morehills branch, then it is a known error in my attempts of fixing the interpolation issue (which is why the morehills has no PR).

PR #997, opened 2017-10-04
hillshading update: filtered upscaling on android, smooth tile borders

usrusr:
Updates to #923 hillshading, the seams reported by  @demiantres should be addressed by this, amongst others.
The biggest changes:
1. Filtered upscaling on android (using a questionably arcane way, but more straight-forward approaches failed to give filtered upscaling on my test Z1c)
2. Perfect (as far as i can tell?) alignment at tile borders 
3. Optional interpolation padding at hillshading tile borders to get rid of interpolation artifacts. Unfortunately it is very slow on android (first map tile rendered after five hgt tiles are loaded instead of after one hgt), it might be faster to just merge the border pixels on the go into a four pixels wide temporary bitmap that is painted separately (as a two-hillpixels wide piece)
4. An "always" attribute in the hillshading theme element that causes flat shading to be applied whenever hgt files are missing or hillshading is not enabled, so that themes can compensate for desaturation

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
@@ -48,5 +48,6 @@ public interface GraphicContext {
 
     void setClipDifference(int left, int top, int width, int height);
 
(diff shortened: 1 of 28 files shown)


devemux86:
Thanks, nice work!

Merged via c691e7197c2daa3cf0663e11c7e7e5dd4147d428.

devemux86:
BTW in Desktop Samples skipped the `-restore` for simplicity, as was saving already the model, so only the check on open was needed.

PR #1003, opened 2017-10-07
Hillshading: shading formula extensions

usrusr:
Finally some updates to the old shading formula prototype:

SimpleShadingAlgorithm is still as arbitrary as ever, but now with generous curve-tweaking to bring out the structure of more subtle landscapes. 

DiffuseShadingAlgorithm is using math instead of making up numbers, it is much better at bringing out the details of mountain ranges and the like (while still showing more in the lowlands than the old SimpleShadingAlgorithm)


devemux86:
Related to #923.

devemux86:
Thanks for the continuous improvements!

Squashed and merged via 347feb990241bd951b78d14822e248110c0d74ac.

PR #1006, opened 2017-10-15
Mapsforge-map: support alterable osm tag values

Gustl22:
With this modification, all OSM-tag values (such as numerical and string values) can be saved in mapsforge .map format without specify a fixed value term in `tag-mapping.xml`. 
For that the wildcards `"%f"` (float) for numerical and `"%s"` for string values can be used. Internally the values are cast to the smallest possible type (like Byte: `%b`, Short: `%h`, Int: `%i`). Using `%s` wildcard is not recommended, better use ready `osm-tag`s. Colors can internally be converted to `int` values with `%f` wildcard.

The map-writer can be used for older versions too, if tag-values are disabled. If not, new specification doesn't allow older readers to read new one. New Readers won't have a problem to read them all. The tag-value bit in file header is not needed, but I set it, because I can't decide if assumes a role for later versions.

The writing performance keeps very similar. I inherited some S3DB-tags, (especially height tag) and the map file size only increases around 0,24% in my case.

_Why is it useful?_ 
Until now, no additional data could be stored (apart from concrete feature tags, like the house number).
Now heights, colors, angles and much more can be inherited, with marginal overhead.
This is not meant to include concrete style elements. E.g. colors and materials are only an assistance for later rendering.

Let me know what you think of ;D

### DIFF
--- a/docs/Specification-Binary-Map-File.md
+++ b/docs/Specification-Binary-Map-File.md
@@ -112,7 +112,7 @@ To read the data of a specific tile in the sub-file, the position of the fixed-s
 |32|yes|POI signature|If the debug bit in the file header is set:<br />`***POIStartX***` where X defines the OSM-ID of the POI; the text is always padded to 32 bytes by adding whitespaces|
 |variable||position|geo coordinate difference to the top-left corner of the current tile as *`VBE-S` INT*, in the order lat-diff, lon-diff|
 |1||special byte|<ul><li>1.-4. bit: layer (OSM-Tag: layer=...) + 5 (to avoid negative values)</li><li>5.-8. bit: amount of tags for the POI</li></ul>|
-|variable||tag id|for each tag of the POI:<ul><li>tag id as *`VBE-U` INT*</li></ul>|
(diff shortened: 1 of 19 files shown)


devemux86:
Thanks, will check it!

- Opening maps < v5, there are rendering issues, like not visible sea.
- Apparently would need relevant changes in [VTM](https://github.com/mapsforge/vtm) reader too.

Gustl22:
- ok. I'll try to fix it.
- An implementation for VTM reader is already finished, if this is considered as good approach.

devemux86:
> ok. I'll try to fix it.

Sorry for the confusion, had opened wrong map file, the sea is rendered fine. :slightly_smiling_face: 

> An implementation for VTM reader is already finished, if this is considered as good approach.

If you have it ready, can post a PR in order to check them in parallel, as Mapsforge cannot render extrusions and would be interesting to see how VTM plays with v5 maps.

devemux86:
Thanks, as usual very nice & clean work! :+1: 

Merged via 7a9b1ca4964282a078704cb3d35033aea12b4cda.

New `tag-values` map-writer option has sane default `false` to allow users explicitly activate v5 maps and not break compatibility with existing apps.

devemux86:
BTW I left out the tag-value bit in file header, map version is written anyway for any needed checks.

Issue #1012, opened 2017-11-06
Mapsforge maps: Root id for building parts


Gustl22:
Hello,
like described in https://mapzen.com/documentation/vector-tiles/layers/#buildings-and-addresses we should add a root id for building parts. So while rendering overlapping buildings can easily be distinguished.

For that id of root elements must be stored. My thoughts:
- only tag buildings with `id`, which inherit parts (preferably OSM-ID)
- give every element its `id` (preferably OSM-ID , more overhead)
- use id inherited order of elements in map-file (_if it is possible_) and add while reading, but seems unsafe
- Id in `---WayStartX---`, but can't be used, only in debug mode

The root id in part elements could be added as follows:
- Use OSM `ref` tag, which normally isn't used otherwise for buildings
- Add an extra `root_id` tag, which does not exist in OSM, but is unique

Correct me, if I interpret something wrong. 
Greets Gustl

devemux86:
Can you explain more the concept for this?
Store extra id for grouping together features?
What are the benefits for the readers, how can be used in rendering?

Gustl22:
Example `height`:
A building with parts can't be distinguished from other buildings, so by default it gets a height or it has the height of highest point of building like described in OSM-conventions ( http://wiki.openstreetmap.org/wiki/Simple_3D_buildings ). Underlaying parts aren't visible and currently there are artefacts when the wall of surrounding building covers its parts.
I tried an implementation which detects the polygons inside a building, but on client side its too cost-intensive. A reference would be much faster and does not take much place.

Features could be adopted from its root element in reader, if necessary.

devemux86:
Could start with the needed elements, like the buildings and use the available `ref` tag to check the implementation and results.
That way map files remain compatible and can see if there is need for special tag in a next step.
Also on client rendering would be interesting to check the results "before" and "after".
Certainly it's best to avoid real time calculations in readers as can hurt performance.

devemux86:
Storing OSM ids in `ref` tag probably could be accomplished with [TagTransform](http://wiki.openstreetmap.org/wiki/Osmosis/TagTransform) Osmosis plugin as a preprocessing step?

Gustl22:
TagTransform may work for already existant relations and tags but i didn't read about calculation out of geo data to create a new relation (reference is predominantly given by the building polygon, which inherits the part).

devemux86:
Fixed via PR #1014 by @Gustl22.

PR #1014, opened 2017-11-16
map-writer maps v5: reference building-parts to buildings (implicit relations)

Gustl22:
Hi,
I added the relation between building parts and its buildings like described in #1012. I called them implicit relations in code. I only added an `id` to the elements which contain parts. Parts have a `ref`-tag.
See [root_id](https://mapzen.com/documentation/vector-tiles/layers/) in Mapzen.
Here an example, how i solved it in VTM; ~~I will start a PR there, if this is a good approach~~:
artifacts

The increasement of file space is not noticeable (only possible with v5). It needs a little bit longer to calculate in HD'processor. 
Under `#HDstoreData` I added a `TODO`, which may can improve the process of whole HD'processing...

While implementing, I noticed that the `tilesToCoastlines` in `BaseTileBasedDataProcessor` wasn't updated by HD'processor, but nothing had influence on the result. What is the reason for that?

BTW: I added some logging code, which is not necessarily bonded on the pull request. You're always welcome to change my code yourself or request changes.

### DIFF
--- a/mapsforge-map-writer/src/main/config/tag-mapping.xml
+++ b/mapsforge-map-writer/src/main/config/tag-mapping.xml
@@ -143,6 +143,9 @@
 
 
     <!-- ************* WAYS *************** -->
+    <ways>
+        <osm-tag key="id" renderable="false" value="%f" />
+    </ways>
 
     <!-- HIGHWAY TAGS -->
     <ways>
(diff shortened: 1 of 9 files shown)


devemux86:
> Here an example, how i solved it in VTM; I will start a PR there, if this is a good approach

Have not looked yet the writer PR, but if need changes also in readers, please post PR for them too, in order to be able to test properly the process in all affected modules.

devemux86:
Thanks, very nice clean work!
Merged via 3e038733efdc2e52c6d68b7b8ec6c6b78b58889a.

PR #1025, opened 2017-12-28
mapsforge-map-writer: reference all part elements

Gustl22:
- reference all part elements (not only the first one; necessary to pass tags on to parts in reader)
- tag-mapping: enable roof directions

Minimal increased size and processing time


devemux86:
Thanks!
Merged via 5e3194f814fea1e9072637687981f7f81ca73a25.

Gustl22:
@devemux86: Is it worth to enable all [directions](https://wiki.openstreetmap.org/wiki/Key:direction) (NNE, NNW etc.) in `tag-mapping`? This would add about 350 ways ([overpass turbo](https://overpass-turbo.eu/s/Bkt)).

Issue #930, opened 2017-03-03
map-writer: infinite loop with polylabel algorithm


Klaus-Tockloth:
In order to improve the positioning of labels like this ...

![bildschirmfoto 2017-03-03 um 19 34 05](https://cloud.githubusercontent.com/assets/5493128/23563888/f5662930-0048-11e7-8a22-2aa8754124ca.jpg)

... I have activated the map-writer option "label-position=true". The result is a lot of warning output and the process runs into an infinite loop and has to be killed.

```
Mar 03, 2017 6:53:51 PM org.openstreetmap.osmosis.core.Osmosis run
INFO: Osmosis Version 0.43.1
Mar 03, 2017 6:53:51 PM org.openstreetmap.osmosis.core.Osmosis run
INFO: Preparing pipeline.
Mar 03, 2017 6:53:51 PM org.mapsforge.map.writer.osmosis.MapFileWriterTask <init>
INFO: mapfile-writer version: mapsforge-map-writer-0.7.0
...
Mar 03, 2017 6:54:26 PM org.mapsforge.map.writer.MapFileWriter writeSubfile
INFO: written 100% of sub file for zoom interval 1 (5,0,7)
Mar 03, 2017 6:54:27 PM org.mapsforge.map.writer.util.PolyLabel get
WARNING: Failed to get label for geometry: MULTILINESTRING ((7.073865 51.483592, 7.073792 51.483496, 7.073848 51.483453, ...
...
Mar 03, 2017 6:54:27 PM org.mapsforge.map.writer.util.PolyLabel get
WARNING: Failed to get label for geometry: MULTILINESTRING ((7.365507 52.29322, 7.364424 52.291585, 7.363985 52.290906, ...
...
```

See also issue #886

devemux86:
From the old days, map writer's label position option is not used in Mapsforge map reader.

That work was initiated from https://github.com/mapsforge/vtm/issues/80 for improved polygon labels and was implemented only in VTM client (not Mapsforge).

cc @andreynovikov

devemux86:
Isn't more safe to use instead the `label-position` in tag-mapping for the needed polygons?
e.g. see default tag-mapping.xml which has it for polygon [buildings](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map-writer/src/main/config/tag-mapping.xml#L312)

Klaus-Tockloth:
Conclusion ... the "label-position=true" feature:
[...]

Issue #891, opened 2016-10-28
Add support for polygon label position


devemux86:
See https://github.com/mapsforge/vtm/issues/80 for related discussion (and implementation in VTM).

Uses the #886 work in map writer.

devemux86:
See [VTM](https://github.com/mapsforge/vtm) OpenGL vector map library, compatible with Mapsforge maps.

Issue #886, opened 2016-10-15
Map writer: polygon label position enhancements


devemux86:
- Implement map-writer configuration per tag for label position
- Improve label position calculation using the algorithm https://www.mapbox.com/blog/polygon-center

devemux86:
See PR #884.

PR #1029, opened 2018-01-19
mapsforge-map-writer: add places of worship as areas

Gustl22:
Until now only nodes are stored as place of worship. So churches and similar not were discernible.
Now areas are displayed too. Also plan to support more ways/areas incl. symbols in VTM.

### DIFF
--- a/mapsforge-map-writer/src/main/config/tag-mapping.xml
+++ b/mapsforge-map-writer/src/main/config/tag-mapping.xml
@@ -280,11 +280,16 @@
         <osm-tag key="amenity" value="grave_yard" zoom-appear="14" />
         <osm-tag key="amenity" value="hospital" zoom-appear="13" />
         <osm-tag key="amenity" value="parking" zoom-appear="14" />
+        <osm-tag key="amenity" value="place_of_worship" zoom-appear="14" />
         <osm-tag key="amenity" value="school" zoom-appear="14" />
         <osm-tag key="amenity" value="swimming_pool" zoom-appear="14" />
         <osm-tag key="amenity" value="theatre" zoom-appear="14" />
         <osm-tag key="amenity" value="toilets" zoom-appear="17" />
         <osm-tag key="amenity" value="university" zoom-appear="13" />
+
+        <osm-tag key="religion" renderable="false" value="christian" />
+        <osm-tag key="religion" renderable="false" value="jewish" />
+        <osm-tag key="religion" renderable="false" value="muslim" />
 
         <osm-tag key="tourism" value="attraction" zoom-appear="14" />
         <osm-tag key="tourism" value="hostel" zoom-appear="15" />


applantation:
Last time I checked there were a couple more religions that should also
make it into this patch.

But on a more general level, there used to a an osmosis plugin area2points
that added a point into the middle of a closed way.
https://wiki.openstreetmap.org/wiki/Osmosis/Detailed_Usage_0.40#--areapoints_.28--a2p.29

This has been, for unknown reasons, dropped from osmosis, but the code
exists and it is simply a question of installing the extension with osmosis
to get it to work.

I think this would be a better way of doing this, as it will mean on the
rendering front we do not require another rule for
churches/temples/mosques/ that appear as closed ways in OSM data.

devemux86:
The religions in the PR are the same as the old ones in <pois> section.
They're been there from start, but eventually all should be improved.

Regarding the plugin seems a nice idea. Though it implies that users will know and use it.
And a perfect map creation should use definitely TagTransform (like OpenAndroMaps).
Here we try to offer just the a default tag-mapping for general purposes.

devemux86:
Thanks!

devemux86:
I updated tag-mapping with more values as exist already in the themes and with available symbols in resources.

devemux86:
Note: in v5 maps such issues should be solved now with variable tag values!

Gustl22:
> _Ludwig:_ But on a more general level, there used to a an osmosis plugin area2points that added a point into the middle of a closed way.

Yes, providers should consider area2points and TagTransform. But then we need a guide, that all can get the same results.

> _Ludwig:_ I think this would be a better way of doing this, as it will mean on the rendering front we do not require another rule for churches/temples/mosques/ that appear as closed ways in OSM data.

No need of duplicate rules for nodes and ways, if we handle each `value` for ways and nodes in one step. Will post a PR in VTM.

> _devemux86:_ Note: in v5 maps such issues should be solved now with variable tag values!

What issues do mean?

devemux86:
> What issues do you mean?

The possible missed values for a tag, can be solved with new variable tag values feature, like we do now in buildings (for v5 maps).

devemux86:
There is already support in map-writer #886 for automatic areas-to-points calculation to be used for labels / symbols. Can be enabled:
- globally via [label-position](https://github.com/mapsforge/mapsforge/blob/master/docs/Getting-Started-Map-Writer.md#advanced-options-only-use-when-you-know-what-you-are-doing) option
- in tag-mapping via [label-position](https://github.com/mapsforge/mapsforge/blob/master/resources/tag-mapping.xsd#L127-L134) attribute

VTM https://github.com/mapsforge/vtm/issues/80 already supports this feature. Not Mapsforge #891.

Though all these need improvements, so are disabled by default.
See issues in map-writer #930 and VTM https://github.com/mapsforge/vtm/issues/402.

PR #1036, opened 2018-02-23
high resolution mapsforge logo

Gustl22:
add mapsforge logo as vector drawable and update `ic_launcher` icon in android samples


devemux86:
Can you add also mapsforge svg without the label?

Gustl22:
Yes, to the ` docs/images/logo` directory then?
If you want I can add Photoshop file (with shadow for Android icons) there, too?

devemux86:
Like the VTM logo svg without label in https://github.com/mapsforge/vtm/pull/508/commits/cbb357c1be16934568a7d0fab45234b9c61e3917, I'll slightly relocate them anyway.

devemux86:
Thanks, nice work!
Squashed and merged via 5744f120541548daa978493794682016889daf8e.

Included only the svg in docs, as much larger eps could not be needed?

Gustl22:
> Included only the svg in docs, as much larger eps could not be needed?

eps is easier to edit (with layers), but not needed ;)

PR #1044, opened 2018-03-15
introducing an interface for MapViewPosition

mikes222:
introducing an interface for MapViewPosition. The alternative implementation of MapViewPosition is already included but not yet fully working with that checkin. In order to get it working we need to adapt the interface a bit. But that is a story for a different pull-request.

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Matrix.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Matrix.java
@@ -27,6 +27,11 @@ public interface Matrix {
      */
     void rotate(float theta, float pivotX, float pivotY);
 
+    /**
+     * Scales around the 0/0 coordinate
+     * @param scaleX the scale factor in x-direction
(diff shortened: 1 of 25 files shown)


devemux86:
Thanks!

Rebased and merged via cd50e018c5f1d829e49d612a6e9d8480f6f92f29.

New class is `MapViewPosition2` but remains a placeholder (extending regular), since it didn't actually work, could not even zoom. So once having a _working_ implementation can be placed in there.

With 70c2778b8a58d50edd8ef561aafcc550ef5303f1 I pushed:
- `MapViewPosition2` can be enabled simply via:
```java
Parameters.MAP_VIEW_POSITION2 = true;
```
- Plus a new `MapViewPosition2Viewer` example in Android samples.

Some general remarks:
- Better use a [topic branch](https://help.github.com/articles/about-pull-requests/) for pull requests for easier maintenance.
- Pull requests must be (re)based on latest master before submit (this one was quite old).

PR #1047, opened 2018-03-30
Manual move/zoom listener

mikes222:


devemux86:
Thanks!
Merged via ecb651881973e7f88e66d23511955c9686c4f5f9 with several improvements, like code simplification, thread-safe `CopyOnWriteArrayList`, more events, etc.

PR #1059, opened 2018-05-06
fix deadlock

drakkan:
If a mapView is created and destroyed very quickly can happen that LayerManager locks a Bitmap using getDrawingBitmap and the lock is not realeased until there is an animation in progress, so PausableThread need to execute all the pending work before exit to avoid deadlocks

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/LayerManager.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/LayerManager.java
@@ -42,7 +42,8 @@ public class LayerManager extends PausableThread implements Redrawer {
 
     public LayerManager(MapView mapView, IMapViewPosition mapViewPosition, GraphicFactory graphicFactory) {
         super();
-
+        
+        this.waitForDoWork = true;
         this.mapView = mapView;
         this.mapViewPosition = mapViewPosition;
 
--- a/mapsforge-map/src/main/java/org/mapsforge/map/util/PausableThread.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/util/PausableThread.java
@@ -123,8 +123,8 @@ public abstract class PausableThread extends Thread {
     public final void run() {
         setName(getClass().getSimpleName());
         setPriority(getThreadPriority().priority);
-		
-        while ((!this.shouldStop && !isInterrupted()) || hasWork()) {
+
+        while (!this.shouldStop && !isInterrupted()) {
             synchronized (this) {
                 while (!this.shouldStop && !isInterrupted() && (this.shouldPause || !hasWork())) {
                     try {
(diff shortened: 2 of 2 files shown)


devemux86:
Thanks!

Unfortunately testing with mapsforge-samples-android now the CPU goes crazy when exit some samples (e.g. LocationOverlayMapViewer) and return to main samples menu screen.

drakkan:
well, the problem is that TileDownloadThread return true hard coded in doWork here

https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/download/TileDownloadThread.java#L76

I don't know why it always return true, we could change this hardcoded value to false or maybe we could change approach, for example we could add a protected member variable for example waitForFinishWork, default false to PausableThread, so any class that extends PausableThread could set to true if needed, LayerManager is actually the only one that need to wait for work to finish to avoid deadlocks, MapViewPosition and TileDownloadThread can be interrupted without waiting for pending job, what do you think about? thanks

devemux86:
That seems much better!

For now I merged it in [issue_1059](https://github.com/mapsforge/mapsforge/tree/issue_1059) so anyone can do further testing.

devemux86:
Thanks!

Merged via 83e6d95a076c044a8045ac2b2c0bc284af38a458.

java1developer:
Bug not fixed for me. Snaphots:
    implementation 'org.mapsforge:mapsforge-themes:master-SNAPSHOT'
    implementation 'org.mapsforge:mapsforge-map-android:master-SNAPSHOT'
MapView in fragment, destroyAll called in onDestroyView
all initialization code (and no more, no layers, no cache and other):
`mapView.visibility = View.GONE`

wait() deadlock in thread
> this = {MapViewPosition$Animator@5834} 
 mapSize = 0
 moveSteps = 0
 scaleDifference = 0.0
 startScaleFactor = 0.0
 targetPixelX = 0.0
 targetPixelY = 0.0
 this$0 = {MapViewPosition@5849} 
 timeEnd = 0
 timeStart = 0
 zoomAnimation = false
 pausing = false
 shouldPause = false
 shouldStop = false
 waitForWork = false
 blocker = null
 blockerLock = {Object@5850} 
 contextClassLoader = {PathClassLoader@5819} 
 daemon = false
 eetop = 0
 group = {ThreadGroup@5839} 
 inheritableThreadLocals = null
 inheritedAccessControlContext = {AccessControlContext@5851} 
 lock = {Object@5852} 
 name = "Animator"
 nativeParkEventPointer = 0
 nativePeer = 500864377344
 parkBlocker = null
 parkState = 1
 priority = 7
 single_step = false
 stackSize = 0
 started = true
 stillborn = false
 target = null
 threadLocalRandomProbe = 0
 threadLocalRandomSecondarySeed = 0
 threadLocalRandomSeed = 0
 threadLocals = null
 threadQ = null
 threadStatus = 0
 tid = 9400
 uncaughtExceptionHandler = null
 shadow$_klass_ = {Class@5725} org.mapsforge.map.model.MapViewPosition$Animator
 shadow$_monitor_ = 1073743835

with variables
`this.waitForWork = false`
`this.pausing = false`
`this.shouldStop = false`

PR #1062, opened 2018-05-16
POI: allow tag keys as category

Gustl22:
I added functionality.
Important is if using key categories, they should be parents of their value categories (if there're any). The key category isn't respected, if any value categories appear. Otherwise an element would be added to the value category and the key category, although it's already a parent...

### DIFF
--- a/mapsforge-poi-writer/src/main/config/poi-mapping.xml
+++ b/mapsforge-poi-writer/src/main/config/poi-mapping.xml
@@ -1546,4 +1546,10 @@
         </category>
     </category>
 
+    <!-- Address search -->
+    <!--<category title="Address">
+        <mapping tag="addr:street" />
+        <mapping tag="addr:housenumber" />
+    </category>-->
+
 </category>
--- a/mapsforge-poi-writer/src/main/java/org/mapsforge/poi/writer/TagMappingResolver.java
+++ b/mapsforge-poi-writer/src/main/java/org/mapsforge/poi/writer/TagMappingResolver.java
@@ -100,7 +100,10 @@ class TagMappingResolver {
 
         // For each mapping's tag: split and save key (uniquely)
         for (String tag : this.tagMap.keySet()) {
-            this.mappingTags.add(tag.split("=")[0]);
+            if (tag.contains("="))
+                this.mappingTags.add(tag.split("=")[0]);
+            else
+                this.mappingTags.add(tag);
         }
     }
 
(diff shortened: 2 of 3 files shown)


devemux86:
Can you mention here an example of the warning (in xml notation) so that users can understand and have a reference?

Gustl22:
Recommended:
```
<category title="Places">
        <category title="Cities">
            <mapping tag="place=city" />
        </category>
        <mapping tag="place" />
</category>
```
Here the city is accessible via `Places` and `Cities` although it only is in `Cities` category.

Not recommended:
```
<category title="Places">
        <category title="Cities">
            <mapping tag="place=city" />
        </category>
</category>
<category title="All places">
        <mapping tag="place" />
</category>
```
In the second case cities are added as category `Places` -> `Cities`, but not as `All places`, even it has the key `place`. 

_Background is that an element shouldn't be added to both categories in the first case (`Places` and `Cities`), but shall further be accessible via parent category._

devemux86:
So mean that in 1st example the `city` is added in value category `Cities`/`Places` and then the rule `<mapping tag="place" />` works for all other `place` values besides the `place=city`?

And in 2nd example the `city` is added 2 times: in value category `Cities`/`Places` and separately in `All places` as that is a discrete parent rule?

mapsforge-poi-writer/src/main/java/org/mapsforge/poi/writer/TagMappingResolver.java
```diff
@@ -100,7 +100,10 @@
 
         // For each mapping's tag: split and save key (uniquely)
         for (String tag : this.tagMap.keySet()) {
-            this.mappingTags.add(tag.split("=")[0]);
+            if (tag.contains("="))
+                this.mappingTags.add(tag.split("=")[0]);
+            else
+                this.mappingTags.add(tag);
```
devemux86: No need for that, split returns the original string if cannot split it.

mapsforge-poi-writer/src/main/config/poi-mapping.xml
```diff
@@ -1546,4 +1546,10 @@
         </category>
     </category>
 
+    <!-- Address search -->
+    <!--<category title="Address">
+        <mapping tag="addr:street" />
+        <mapping tag="addr:housenumber" />
+    </category>-->
```
devemux86: We could have them enabled for broader visibility and test of this new feature.

devemux86:
Thanks!

Merged via c4c0c0ec3cff90f5cdb4af874e8e3785540e1da5.

PR #1067, opened 2018-06-14
Map writer: improve progress logs

Gustl22:
Especially for HD processing.


devemux86:
Thanks!
Merged via 460d1a92784886faf490c00da9441207bc25380d.

devemux86:
Further improving map + poi writers, we can include an extra option to these progress logs.
That's useful if build on server to not fill its log with unneeded multiple progress lines.

PR #1068, opened 2018-06-21
adding mapsforge-ovtencoder subproject

ikroshlab:
This subproject aims to have a possibility of encoding map tiles from Mapsforge map files and deploy them to an online tile server.


devemux86:
We can review it with an improved implementation with the above notes.

PR #1070, opened 2018-06-24
fix #1069 NPE in ClusterManager.getCurBounds()

### LINKED ISSUE #1069: Possible NullpointerException in ClusterManager.getCurBounds
I stumpled upon a NullpointerException in the Android samples ClusterManager, like reported here:
https://groups.google.com/forum/#!topic/mapsforge-dev/kk2PRQbwSWw

Happening here: https://github.com/mapsforge/mapsforge/blob/51a3350d21812cda25cebcbac9cb5dcf4e79d0f5/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/cluster/ClusterManager.java#L293

Looks like it's caused when the BoundingBox is not in the current viewport.

I've a small patch to handle possible null here. I'll send a PR with it.

pstorch:
The local variables nw_ and se_ might be null, if the BoundingBox is not in the current viewport.
This patch adds a null check.


devemux86:
Thanks!
Merged via 369de625e28b6be1e1969d271c36a643462762ad.

PR #1073, opened 2018-06-26
Map writer: adjust implicit way relations

Gustl22:
Now all parts of buildings should be related. Not only these ones which are completely contained.

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/BaseTileBasedDataProcessor.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/BaseTileBasedDataProcessor.java
@@ -645,8 +645,12 @@ abstract class BaseTileBasedDataProcessor implements TileBasedDataProcessor, Nod
                         }
                     }
 
-                    // Memorize multi-polygons
-                    if (rPolygon.contains(pPolygon)) {
+                    /* Use interior point cause
+                     * - boundary points may intersect with other geometries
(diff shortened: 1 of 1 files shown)


devemux86:
Thanks!

mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/BaseTileBasedDataProcessor.java
```diff
@@ -645,8 +645,12 @@ protected void handleImplicitWayRelations() {
                         }
                     }
 
-                    // Memorize multi-polygons
-                    if (rPolygon.contains(pPolygon)) {
+                    /* Use interior point cause
+                     * - boundary points may intersect with other geometries
+                     * - part polygon often not completely covered by root polygon
+                     */
+                    if (rPolygon.covers(pPolygon.getInteriorPoint())) {
```
devemux86: Or "intersects" could work too?

Gustl22: Not good enough, cause of this:

> boundary points may intersect with other geometries

This means that part boundary can intersect (cause of imprecision) with another independent building (which is directly drawn aside), but in fact it belongs to another building. So it would be assigned (depending on the process order) to the wrong building.
Using an interior point ensures that it is referenced to the building whose area "covers" it the most.

PR #1087, opened 2018-09-10
tag-mapping: add potential 3d POIs

Gustl22:
Increases file size around 1 percent in densely populated areas. Elsewhere should be less.


devemux86:
Thanks!

PR #1091, opened 2018-10-03
Render themes: PNG scaling

b3nn0:
As reported yesterday, here is my proposed implementation for DPI scaling and respecting symbol-width/height/percent.
It also contains a small refactoring that puts the code that determines the final image size from the given parameters in a separate function. Previousely, the exact same code existed twice for Android and AWT. Now it exists only once in mapsforge-core/GraphicUtils.


mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidResourceBitmap.java
```diff
@@ -82,6 +83,9 @@ public static void clearResourceBitmaps() {
                 if (bitmap == null) {
                     throw new IOException("BitmapFactory failed to decodeStream");
                 }
+                float[] finalSize = GraphicUtils.computeFinalImageSize(bitmap.getWidth(), bitmap.getHeight(), scaleFactor, width, height, percent);
+                if ((int) finalSize[0] != width || (int) finalSize[1] != height)
```
devemux86: The comparison should be done with bitmap's width and height.

b3nn0: Whoops, you are right. Fixed.

devemux86:
Thanks!
Merged via 62e0875acea039a74426442562e65c6d895b0d60.

PR #1100, opened 2019-02-14
Don't draw cached parent tile

cpt1gl0:
Drawing the cached parent tile if the child is not available makes the UI very irresponsive in some situations. You might provoke this by  zooming into the map and scroll the map before the new tiles are loaded (I was using a TileDownloadLayer for this). In general the tile layer seems much more responsive without drawing cached parents. Give it a try if you don't believe it, you will notice the difference.


cpt1gl0:
I investigated a bit further and found out that the canvas.draw call for the cached parent is blocking for very long eventually.

devemux86:
That leads to empty map on each zoom, not what users would expect from map rendering.

We can examine it as option via `Parameters` class, so not break users existing workflow.

cpt1gl0:
I now found a way to make parent tile drawing faster (see #1101).

devemux86:
So we can close here.

PR #1101, opened 2019-02-16
Faster parent tile drawing

cpt1gl0:
Make parent tile drawing faster by using Android canvas draw bitmap method for transforming the parent tile.

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
@@ -28,8 +28,7 @@ public interface GraphicContext {
 
     void drawBitmap(Bitmap bitmap, Matrix matrix, Filter filter);
 
-    void drawBitmap(Bitmap bitmap, Rectangle src, Rectangle dst,
-                    Filter filter, boolean fastScaling);
+    void drawBitmap(Bitmap bitmap, Rectangle src, Rectangle dst, Filter filter);
 
     void drawCircle(int x, int y, int radius, Paint paint);
 
@@ -50,6 +49,9 @@ public interface GraphicContext {
     void setClip(int left, int top, int width, int height);
 
     void setClipDifference(int left, int top, int width, int height);
+
+    InterpolationMode getInterpolationMode();
+    void setInterpolationMode(InterpolationMode mode);
 
     /**
      * Shade whole map tile when tileRect is null (and bitmap, shadeRect are null).
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
@@ -28,7 +28,8 @@ public interface GraphicContext {
 
     void drawBitmap(Bitmap bitmap, Matrix matrix, Filter filter);
 
-    void drawBitmap(Bitmap bitmap, Rectangle src, Rectangle dst, Filter filter);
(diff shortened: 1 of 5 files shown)


cpt1gl0:
I think the only thing missing is an implementation of the newly introduced bitmap draw method in AWT canvas.

devemux86:
> I think the only thing missing is an implementation of the newly introduced bitmap draw method in AWT canvas.

Indeed, or else the library cannot be built and should work in all supported platforms.

devemux86:
Thanks, from some quick tests on Android emulator cannot see any _clear_ performance change?
Need to implement on Desktop too, so we can test easier the changes there.

Also `TileLayer` cannot be compiled, it has missing imports.

cpt1gl0:
I added the missing includes and implemented srd/dst bitmap drawing for the AWT canvas.

devemux86:
Thanks for the fixes. Still cannot see any notable performance benefit?
But there is regression in parent tiles rendering, now there are seams among them.
That comes because of the loss of precision in new draw methods (casting to `int`).

devemux86:
tiles

cpt1gl0:
I also hardly notice any difference on my mobile phone, but on the phone of my girlfriend it really gets irresponsive for ~1 second. I think it might be a memory issue. I guess the matrix transformation needs more memory than the simple just copying pixels from a certain area.
I am not sure if the seams are because of the rounding. Could you send me, or point me to some example code where I can easily see them?

devemux86:
Cannot see anything with current `master`, though with the PR the change is immediately seen with first quick tests.

Like mentioned in previous PR we can add an option in `Parameters` for anyone wanting to skip parent tiles rendering if that's the way for speed increase.

cpt1gl0:
I think the reason for the seams is the interpolation done when scaling the sub-regions of the parent tile. For the original approach there are no seams because the parent tile is scaled as a whole.

In the latest commit I disabled anti-aliasing and bilinear interpolation for parent tile scaling and would expect the seams to be gone now. - It would be great if you could check again. In addition, with interpolation disabled for parent tile drawing I notice another slight increase of rendering performance on my phone.

devemux86:
Disabling such filters makes rendering worse / pixelated - not wanted for many users - for possibly gain a little performance on some devices. Cannot force such default behavior to all users.

If don't want parent tiles rendering, best is to make it optional, so can users are free to select their way.

devemux86:
For fast rendering with extra features like rotation / tilt, better see an OpenGL map library like [VTM](https://github.com/mapsforge/vtm).

cpt1gl0:
Would you be fine with having it optional so users can choose between "no parent tile rendering", "fast parent tile rendering" and "full parent tile rendering"?

devemux86:
Simplest solution is an on / off switch, for anyone needing that workflow.
Pushed `Parameters.PARENT_TILES_RENDERING` option via 08db3cfe344dcfe53cbd08e5c3cbcd9c1ecacd01.

cpt1gl0:
That's for sure simpler and good to have. Still, I'd prefer to have "fast" parent tiles instead of no parent tiles at all. Would you be fine with an additional parameter FAST_PARENT_TILES_RENDERING to enable the fast rendering algorithm?

devemux86:
Implementation will become more complicated for minimum benefit (still cannot see any greater performance).

We can accelerate the whole map rendering if disable every possible anti-aliasing everywhere in lib, that does not solve the fact that some devices have issues, the solution is [accelerated rendering](https://github.com/mapsforge/mapsforge/pull/1101#issuecomment-464627019), not reduce map quality.

PR #1102, opened 2019-02-19
Optional fast parent tiles

cpt1gl0:
Maybe you could reconsider my pull request: I added another parameter FAST_PARENT_TILES_RENDERING to enable the fast rendering algorithm optionally. Modifications to TileLayer itself are minimal with just a small additional if-block for fast rendering.

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/GraphicContext.java
@@ -28,6 +28,8 @@ public interface GraphicContext {
 
     void drawBitmap(Bitmap bitmap, Matrix matrix, Filter filter);
 
+    void drawBitmap(Bitmap bitmap, Rectangle src, Rectangle dst, Filter filter);
+
     void drawCircle(int x, int y, int radius, Paint paint);
 
     void drawLine(int x1, int y1, int x2, int y2, Paint paint);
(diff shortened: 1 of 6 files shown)


devemux86:
Thanks, merged as simplified implementation via 291a31c74312883072a70b981f1989d298c0a6ff.

Also no need for extra draw methods with less precision since play with rendering quality.

_Note: for [contributions](https://github.com/mapsforge/mapsforge/blob/master/.github/CONTRIBUTING.md) please try to follow the conventions you already see in the code._

cpt1gl0:
This will speed up rendering, but not drastically reduce the amount of required memory which comes from scaling not only a region but the entire parent tile bitmap to the desired size, as is done with the matrix transformation.

devemux86:
Already over-solved parent tiles rendering without some evidence for either speed or memory.
If they are so much trouble for some devices, can always disable them with available options.

devemux86:
We can proceed with a different draw on speed rendering mode, but not with so many rectangle objects created for each and every tile continuously.
Can use instead detailed parameters like in Desktop native method `(srcLeft, srcTop, srcRight, srcBottom, srcLeft, srcTop, srcRight, srcBottom)`.

devemux86:
Merged an improved implementation via 8245d0f86b96eb1a9481e14230661042a0b5c722.

cpt1gl0:
> Merged an improved implementation via [8245d0f](https://github.com/mapsforge/mapsforge/commit/8245d0f86b96eb1a9481e14230661042a0b5c722).

That's perfect, thank you. I also prefer your modified implementation with passing single bitmap positions instead of src/dst rectangles to the draw method. It makes clear that positions are expected to be integers.

I am looking forward to the next release including "speed parent tile rendering". Are you planning to release any time soon?

devemux86:
> Are you planning to release any time soon?

There is not currently any fixed date.
I usually release both Mapsforge + VTM simultaneously (as they share offline map progress).

It's easy to use the [snapshot](https://github.com/mapsforge/mapsforge/blob/master/docs/Integration.md#snapshots) releases, always keeping them in good state.
(that's why I'm strict in PR reviews :slightly_smiling_face: )

cpt1gl0:
I see, but I would like to link against a specific version of mapsforge, so there are no surprises if the snapshot is updated. I think this is not possible with the provided snapshots.

PR #1111, opened 2019-03-14
map-writer: ensure that implicit relations can be written (with id tag) for v5 maps

Gustl22:
Ensure that implicit relations can be processed, even if "id" tag is not present in "tag-mapping" file. See [forum](https://groups.google.com/forum/#!topic/mapsforge-dev/Z2Bl4exwWWU).


devemux86:
Thanks, merged via d97eed0d10348a9d8de50fdb1078a1c834f027eb.

PR #1112, opened 2019-03-28
Curved pathText rendering, increased number of road names (#885)

### LINKED ISSUE #885: Rendered result only displays big cities names and no streets/roads
First, allow me to thank you for your amazing work!

I spotted this issue, not sure what the source of the problem is, but in the hope that it will be helpful to you, here are the steps to reproduce it:
- Download the full mapsforge project from github (from October 14th, 2016)
- Download the "brazil.map" http://ftp-stud.hs-esslingen.de/pub/Mirrors/download.mapsforge.org/maps/south-america/brazil.map the version says "2016-10-15 02:56 628M"
- Rename the "brazil.map" to "germany.map" at the root of the sd card
- Open the mapsforge project with Android studio
- Launch the sample app on a real physical device running Android 6.0.1 (LG G5 in my case)
- open the "rendertheme V4" activity

=> when zooming on any city, the streets/roads to not display

b3nn0:
Hi,
first try of an implementation to fix #885.
Current remaining issues:
- TextStroke is taken from here: http://www.jhlabs.com/java/java2d/strokes/
Interestingly, the websites title is "Fun with Java2D". Weird. I can't feel the fun at all ;)
It does currently not obey the MapsForge coding style. Also I have changed two or three lines in the file to make it work correctly for this purpose. How should I go about this?...
- Sometimes, there is some label collision happening. E.g. for multi-lane highways. No idea why:
grafik
- Unit tests. I can't seem to get them working. For Android, I use Android Studio. For the AWT version no idea what to use. I'm debugging with VSCode, but it somehow can't find JUnit, so I can't run the unit tests. How to do a proper dev environment setup for AWT?

Other than that, it looks quite decent. Finally, one can also read some street names in the lower zoom levels. Personally, I'd even go one lower with the min-levels in the default render theme (but let them repeat less often).

Take this as an RFC, not a final patch.

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/model/LineString.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/model/LineString.java
@@ -1,0 +1,109 @@
+package org.mapsforge.core.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class LineString {
+    private List<LineSegment> segments = new ArrayList<>();
+
+    public void LineString() {
+    }
+
+    public void addSegment(LineSegment segment) {
+        this.segments.add(segment);
+    }
+
+    public LineSegment getSegment(int index) {
+        return segments.get(index);
+    }
+
+    public double length() {
+        double res = 0;
+        for (LineSegment segment : segments)
+            res += segment.length();
+        return res;
+    }
+
+    public int segmentCount() {
+        return segments.size();
+    }
+
+    /**
+     * Interpolates on the segment and returns the coordinate of the interpolation result.
+     * Returns null if distance is < 0 or > length()
+     * @param distance
+     * @return
+     */
+    public Point interpolate(double distance) {
+        if (distance < 0)
+            return null;
+
+        for (LineSegment seg : segments) {
+            double segLen = seg.length();
+            if (distance < segLen)
+                return seg.pointAlongLineSegment(distance);
+            distance -= segLen;
+        }
+        return null;
+    }
+
+    /** Creates a new LineString that consists of only the part between startDist and endDist
+    */
+    public LineString extractPart(double startDist, double endDist) {
+        LineString res = new LineString();
+
+        for (int i = 0; i < segments.size(); startDist -= segments.get(i).length(), endDist -= segments.get(i).length(), i++) {
+            LineSegment seg = segments.get(i);
+
+            // Skip first segments that we don't need
+            double segLen = seg.length();
+            if (segLen < startDist)
+                continue;
+
+            Point startPoint = null, endPoint = null;
+            if (startDist >= 0) {
+                // This will be our starting point
+                startPoint = seg.pointAlongLineSegment(startDist);
+            }
+            if (endDist < segLen) {
+                // this will be our ending point
+                endPoint = seg.pointAlongLineSegment(endDist);
+            }
+
+            if (startPoint != null && endPoint == null) {
+                // This ist the starting segment, end will come in a later segment
+                res.addSegment(new LineSegment(startPoint, seg.end));
+            } else if (startPoint == null && endPoint == null) {
+                // Center segment between start and end segment. Add completely
+                res.addSegment(seg);
+            } else if (startPoint == null && endPoint != null) {
+                // End segment, start was in earlier segment
+                res.addSegment(new LineSegment(seg.start, endPoint));
+            } else if (startPoint != null && endPoint != null) {
+                // Start and end on same segment
+                res.addSegment(new LineSegment(startPoint, endPoint));
+            }
+
+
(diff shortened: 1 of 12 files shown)


devemux86:
Thanks, will review it.

> Unit tests

Mapsforge development can be done entirely inside Android Studio for all platforms.
Also tests exist already in several modules, like mapsforge-map, mapsforge-map-awt, etc.

b3nn0:
Ahh, thanks for the hint with Android Studio. Now it's working fine and I can run the test.
Let me know what you think of the general approach / code layout. When this is somewhat final, I will start writing a couple of unit tests. especially for the LineString class that seems to be required.

devemux86:
_About pull requests, it's recommended to use a topic branch, instead of working locally on master branch. See also [GitHub documentation](https://help.github.com/en/articles/about-pull-requests) for more details._

Regarding cut labels, Mapsforge already handles gracefully labels & symbols at tile boundaries, either with regular rendering or with label layer.
So if in pull request's code now it's not working anymore, then something broke library's correct behavior and that must not happen.
Also such large scale factor is of no practical use without proper increase of tile size too.

b3nn0:
About separate branch: Yeah I know. I've fallen into that trap before ;) However, I'm not planning any other mapsforge development right now, so I think it's fine.

I've just pushed another set of changes.
- Unit test for the new LineString class
- Fixed the broken and overlapping labels. It was due to a typo in the bounding box computation.

I would now consider the technical side pretty much done and ready to be merged. If you don't mind, I will do some more changes to the default and osmarender themes to reduce the number of labels a bit.
Now that we don't rely on long line segments any more, there are quite a lot of labels now, so increasing the repeat-gap a bit seems reasonable.

One more screenshot, just for the nice looks:
image

mapsforge-core/src/main/java/org/mapsforge/core/model/LineString.java
```diff
@@ -0,0 +1,126 @@
+package org.mapsforge.core.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class LineString {
+    private List<LineSegment> segments = new ArrayList<>();
```
devemux86: Can expose `segments` and avoid implementing many access methods.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPath.java
```diff
@@ -59,8 +61,13 @@ public void moveTo(float x, float y) {
...
 
+
     @Override
     public void setFillRule(FillRule fillRule) {
         this.path2D.setWindingRule(getWindingRule(fillRule));
     }
+
+    Path2D getRawPath() {
+        return this.path2D;
+    }
```
devemux86: `path2D` is already accessible where is necessary.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/TextStroke.java
```diff
@@ -0,0 +1,161 @@
...
+                            float px = (float) p.getX();
+                            float py = (float) p.getY();
+                            float x = lastX + next * dx * r;
+                            float y = lastY + next * dy * r;
+                            float advance = nextAdvance;
+                            nextAdvance = currentChar < length - 1 ? glyphVector.getGlyphMetrics(currentChar + 1).getAdvance() * 0.5f : 0;
+                            t.setToTranslation(x, y);
+                            t.rotate(angle);
+                            t.translate(-px - advance, -py);
+                            t.translate(0, glyphVector.getVisualBounds().getHeight() / 2); // move slightly down to
```
devemux86: Labels are still not centered, because of way text container was made larger by text height.

devemux86:
Thanks, very nice work!

Squashed and merged via d5359eb716df92c309d440cf607305cd53e07d1a, with format and sort methods improvements.

I pushed via 516dd6c83e7ec580c41418cc450fa926ffba65a0 the fix for proper center the labels also on desktop.

devemux86:
> Now that we don't rely on long line segments any more, there are quite a lot of labels now, so increasing the repeat-gap a bit seems reasonable.

That could be handled:
- Either separately in each theme xml file
- Or change the defaults centrally in [PathText](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/PathText.java#L43-L44) and [renderTheme.xsd](https://github.com/mapsforge/mapsforge/blob/master/resources/renderTheme.xsd#L235-L236)

devemux86:
I changed repeat gap to 100 and we'll see how it appears.

PR #1125, opened 2019-07-03
Get rid of max aspect ratio check in frame buffer controller

cpt1gl0:
Got rid of max aspect ratio check in frame buffer controller to avoid issues with large aspect ratio devices and also because the check seems to be unnecessary.


devemux86:
Thanks, merged via ac6076f77f1129c26a4f1ff8eef55e24765b463c.

PR #1131, opened 2019-09-25
Fix for hill shading issues on Android API 28+

Sublimis:
Canvas.save() and restore() calls (being made on Android API 28 and up) cause intermittent hill shading issues: After zooming the map in/out a few times, some tiles (somewhat randomly chosen) loose their hill shading. Tested on Android 6 (which had no problems), 8, and 9. Removing those calls fixes the issue, and apparently without any side-effects.

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidCanvas.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidCanvas.java
@@ -263,13 +263,13 @@ class AndroidCanvas implements Canvas {
 
     @Override
     public void resetClip() {
-        // if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-        //     this.canvas.save();
-        //     this.canvas.clipRect(0, 0, getWidth(), getHeight());
-        //     this.canvas.restore();
-        // } else {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            // this.canvas.save();
+            this.canvas.clipRect(0, 0, getWidth(), getHeight());
+            // this.canvas.restore();
+        } else {
             this.canvas.clipRect(0, 0, getWidth(), getHeight(), Region.Op.REPLACE);
-        // }
+        }
     }
 
     @Override
@@ -284,13 +284,13 @@ class AndroidCanvas implements Canvas {
 
     @Override
     public void setClip(int left, int top, int width, int height) {
-        // if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-        //     this.canvas.save();
-        //     this.canvas.clipRect(left, top, left + width, top + height);
-        //     this.canvas.restore();
-        // } else {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            // this.canvas.save();
+            this.canvas.clipRect(left, top, left + width, top + height);
+            // this.canvas.restore();
+        } else {
             this.setClipInternal(left, top, width, height, Region.Op.REPLACE);
-        // }
+        }
     }
 
     @Override
@@ -320,13 +320,13 @@ class AndroidCanvas implements Canvas {
 
         if (bitmap == null) {
             if (tileRect != null) {
-                // if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                //     this.canvas.save();
-                //     this.canvas.clipRect((float) tileRect.left, (float) tileRect.top, (float) tileRect.right, (float) tileRect.bottom);
-                //     this.canvas.restore();
-                // } else {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    // this.canvas.save();
+                    this.canvas.clipRect((float) tileRect.left, (float) tileRect.top, (float) tileRect.right, (float) tileRect.bottom);
+                    // this.canvas.restore();
+                } else {
                     this.canvas.clipRect((float) tileRect.left, (float) tileRect.top, (float) tileRect.right, (float) tileRect.bottom, Region.Op.REPLACE);
-                // }
+                }
             }
             // scale a dummy pixel over the canvas - just drawing a paint would probably be faster, but the resulting colors can be inconsistent with the bitmap draw (maybe only on some devices?)
             this.canvas.drawBitmap(hillshadingTemps.useNeutralShadingPixel(), hillshadingTemps.useAsr(0, 0, 1, 1), hillshadingTemps.useAdr(0, 0, canvas.getWidth(), canvas.getHeight()), shadePaint);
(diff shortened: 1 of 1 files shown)


devemux86:
[Canvas.clipRect(float, float, float, float, Region.Op)](https://developer.android.com/reference/android/graphics/Canvas.html#clipRect(float,%20float,%20float,%20float,%20android.graphics.Region.Op)) was deprecated in API level 26.

Instead of revert [Android 9 compatibility](https://github.com/mapsforge/mapsforge/commit/b7b07619a6389886734acd5bcb8ae9b14dfce8c2#diff-015efcf94dedf8aa6199e4c6b84d5e14), better fix any issues with new Android API.

devemux86:
> Sorry, my IDE never warned me about deprecation. Now I simply removed only calls to canvas.save() and restore(), which also fixes the issue for me (again apparently without side-effects). If this simple workaround is not acceptable for other reasons, I think much more complicated and deeper code changes would be needed to address this.

Probably, since the commented lines make worse the vector map rendering, e.g. when zoom in.

devemux86:
> Sorry, my IDE never warned me about deprecation. Now I simply removed only calls to canvas.save() and restore(), which also fixes the issue for me (again apparently without side-effects). If this simple workaround is not acceptable for other reasons, I think much more complicated and deeper code changes would be needed to address this.

Something else is needed or changes in hillshading.
The removed lines break map rendering, e.g. zoom in.

Can find more details for the deprecated method online: https://stackoverflow.com/a/50247323

Sublimis:
Thanks for the link, in fact saw it prior to posting. Problem with the current code is that the following is basically a no-op instruction:

`canvas.save()
canvas.clipRect(...)
canvas.restore()
`

It has no effect whatsoever: restore() operation immediately reverts any effect that clipRect(...) might have had.

In any case, the proposed workaround (removing canvas.save() and canvas.restore() calls from the code like the example above) seems to work for us. We're not using org.mapsforge.map.android.view.MapView, but instead rely on 3rd party library to draw the tiles so this might be the reason. Maybe this will help someone.

devemux86:
That is what was needed so that vector map rendering works as before in Mapsforge.
Cannot comment for third-party rendering, as it's out of library's interest and workflow.
Every test needs be done internally first. Does [HillshadingMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/HillshadingMapViewer.java) sample works?

devemux86:
I enhanced the graphics API and fixed it via different calls for hillshading via 6ff0d3f4856195aa327e1100984e20e3835b62f3.

Sublimis:
Thanks for your efforts! But, not sure we're on the same page. Most important thing to notice is that this:

```java
    @Override
    public void setClip(int left, int top, int width, int height, boolean save) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            if (save) {
                this.canvas.save();
            }
            this.canvas.clipRect(left, top, left + width, top + height);
            if (save) {
                this.canvas.restore();
            }
        } else {
            this.setClipInternal(left, top, width, height, Region.Op.REPLACE);
        }
    }
```

is semantically equivalent to this:

```java
    @Override
    public void setClip(int left, int top, int width, int height, boolean save) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            if (save) {
                // DO NOTHING!
            }
            else {
               this.canvas.clipRect(left, top, left + width, top + height);
            }
        } else {
            this.setClipInternal(left, top, width, height, Region.Op.REPLACE);
        }
    }
```

So basically when the `save` flag is `true`, nothing happens on API 26+. More clarification here: https://stackoverflow.com/questions/29040064/save-canvas-then-restore-why-is-that.

The easiest solution to make clip bounds enlarging possible would be to save original clip bounds when Canvas is created, then restore them back as needed. For this to work, `setBitmap()` method shouldn't be part of `org.mapsforge.core.graphics.Canvas` interface. Instead, new canvas should be created when bitmap change is needed. Also, the no-parameter constructor shouldn't exist, and only the one that accepts Bitmap should be present.

If you can get rid of the aforementioned `org.mapsforge.core.graphics.Canvas.setBitmap()` method and the no-parameter constructor, we can do the rest of coding to implement this properly.

A sidenote, our project uses mapsforge to render tiles, of course. Third-party library is used only to show those tiles (as bitmaps) on the screen.

devemux86:
> is semantically equivalent to this:

No, it's not, how could be?
2nd breaks the zoom, [mentioned that](https://github.com/mapsforge/mapsforge/pull/1131#issuecomment-535126907), have you tested it?
Also cannot understand what you mean with all other _new_ changes.

Hillshading was reporting not working, was confirmed with HillshadingMapViewer sample on Android 8+ and was fixed.

> A sidenote, our project uses mapsforge to render tiles, of course.

That's not mapsforge purpose.
If there is something to help with the public API we can examine it, but we cannot rewrite the library with that purpose.

Sublimis:
> No, it's not, how could be?

Please check the link provided in our last reply. `Restore` reverts `clipRect`.

devemux86:
So you propose at every `clipRect` with `Region.Op.REPLACE` on Android 8+ to not call anything?

Sublimis:
> 
> 
> So you propose at every `clipRect` with `Region.Op.REPLACE` on Android 8+ to not call anything?

No; on Android 1 to 25 this works as intended, as `Region.Op.REPLACE` is supported there. I'm talking about Android 26+. The code above doesn't have **any** effect on Android 26+ whatsoever. The code first checks that it's running on Android 26+; when it is (first branch of the `if`) and the `save` flag is true, it saves the clip, modifies the clip, then immediately restores the clip. This is a no-op.

devemux86:
Mean change like in 4b543dc07b12cd35d3471615f5d66084273f2b45?

devemux86:
Mean changes like in [branch hillshade](https://github.com/mapsforge/mapsforge/tree/hillshade)?

Sublimis:
> 
> 
> Mean change like in [4b543dc](https://github.com/mapsforge/mapsforge/commit/4b543dc07b12cd35d3471615f5d66084273f2b45)?

This change just simplifies the code, but unfortunately it behaves exactly the same. The intention was for you to see that the whole approach is (most probably) futile on Android 26+. Even if it works now, it will give undesirable effects as soon as someone (i.e. rest of the library code) tries to _enlarge_ clip bounds, or calls `AndroidCanvas.setBitmap()` with bitmap of different dimensions. Can you somehow assert that this won't happen?

devemux86:
We can only test and support via the mapsforge-samples-android (or desktop).
Hillshading was verified there with issues on Android 8+ and should work now.

For anything more related to API, can explain in detail in [forum](https://groups.google.com/group/mapsforge-dev) and see if improvements can be made without changing library behavior.

Sublimis:
Ok, thank you for now. We'll see some more what we can do about it in the future!

devemux86:
I merged the changes in master via 90be199ddec956358589cbd0f031ca737e3907cc.

PR #1134, opened 2019-11-02
add removeAll to speed up deletion of larger amounts of layers

cpesch:
addAll exists but no removeAll - is quite a nice speed up for 30k layers


devemux86:
Thanks, merged via 5324d9bf954137f8d75c447e39abe34f0edb72bc with the needed layer group support via 946499b6f3f0f17752ba6e0a8202cc74b755e408.

PR #1136, opened 2019-11-03
Layers: return result of addAll operation of underlying java.util.List

cpesch:
I'm debugging strange effects on AWT and knowing that all layers have been added is helpful.

I don't understand the groups feature, so I didn't touch it.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/Layers.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/Layers.java
@@ -171,8 +171,8 @@ public class Layers implements Iterable<Layer>, RandomAccess {
      * @param layers The new layers to add
      * @see List#addAll(Collection)
      */
-    public synchronized void addAll(Collection<Layer> layers) {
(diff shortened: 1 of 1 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/layer/Layers.java
```diff
@@ -111,15 +111,18 @@ public synchronized void add(Layer layer) {
...
-    public synchronized void add(Layer layer, boolean redraw) {
+    public synchronized boolean add(Layer layer, boolean redraw) {
         checkIsNull(layer);
         layer.setDisplayModel(this.displayModel);
 
-        this.layersList.add(layer);
-        layer.assign(this.redrawer);
-        if (redraw) {
-            this.redrawer.redrawLayers();
+        if(this.layersList.add(layer)) {
```
devemux86: This is always true.

devemux86:
Thanks, merged via be1fccf89ebf30b208166e999d430a30565431a8 (including layer groups).

PR #1140, opened 2019-12-23
automatic scaling to DisplayModel tileSize for tiles from TileStore

mg4gh:
see discussion:
https://groups.google.com/forum/#!topic/mapsforge-dev/AH8nRKgwGDQ


devemux86:
Those are 2 different features and like were discussed in forum, should be posted separately.

Can you create one pull request per feature?

PR #1143, opened 2019-12-23
2nd attempt: Multiple map layers with transparency (configurable alph…

mg4gh:
…a value per Map layer)

see discussion on
https://groups.google.com/forum/#!topic/mapsforge-dev/3ni1635KFmo


devemux86:
Thanks, merged via 634cea3130220e5303f31fb8cd6bde7583074e5c.

Added `alpha` parameter in current graphics API to avoid too many methods.

PR #1145, opened 2020-01-08
Remove OSMapTuner which appears to be dead

TomasHubelbauer:
The URL doesn't work anymore and I didn't find an alternative one that worked.


devemux86:
Several apps may not be active from time to time, but there is no need to remove them.

PR #1149, opened 2020-02-18
Fix: DPI auto scaling doesn't respect rendertheme's symbol-width and symbol-height

sudokai:
DPI scaling is already performed by GraphicUtils::imageSize.
Symbol.width and Symbol.height should just mirror their XML values.
Otherwise, setting explicit symbol-width and symbol-height values
is useless, because PNGs still get scaled nonetheless.

Related commit: b111d3b

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Symbol.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Symbol.java
@@ -73,13 +73,13 @@ public class Symbol extends RenderInstruction {
             } else if (PRIORITY.equals(name)) {
                 this.priority = Integer.parseInt(value);
             } else if (SYMBOL_HEIGHT.equals(name)) {
-                this.height = XmlUtils.parseNonNegativeInteger(name, value) * displayModel.getScaleFactor();
+                this.height = XmlUtils.parseNonNegativeInteger(name, value);
             } else if (SYMBOL_PERCENT.equals(name)) {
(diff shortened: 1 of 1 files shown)


devemux86:
That would break library behavior, as users expect all theme sized elements to be scaled based on dpi.

If something like that is absolutely needed, could be optional, e.g. via `Parameters` class.

sudokai:
I see. I'm trying to display PNG symbols on a map and I have the same PNG symbols in 3 sizes (specified 3 separate render themes). Then depending on the user's display density, I load one of the 3 render themes. I'm trying to get the symbols to look as uniformly sized and as sharp as possible on all kinds of displays.

What would be the best practice here? I'm quite new to the mapsforge project, so apologies for the dumb questions.

devemux86:
Generally PNG are not suitable for mobile devices, due to the vast amount of different screen sizes and dpi.

> I'm trying to display PNG symbols on a map and I have the same PNG symbols in 3 sizes (specified 3 separate render themes). Then depending on the user's display density, I load one of the 3 render themes.

That's a very old (and never perfect) way to handle themes and a waste of resources in development and maintenance.
It was common some years ago before we introduce the SVG resources and their dynamic sizing based on screen dpi.

Clearly the best advice is to use SVG resources everywhere, like we already do in library themes.

There is a huge collection of SVG symbols online and here in the [repository](https://github.com/mapsforge/mapsforge/tree/master/mapsforge-themes/icons/symbols) if want to check it.

devemux86:
Let's continue the discussion in the [forum](https://groups.google.com/group/mapsforge-dev).

sudokai:
Thanks!

Issue #1151, opened 2020-02-26
Persistent tile cache invalidation should happen automatically if tile size has changed.


sudokai:
Some Android phones such as Huawei Mate 20 have dynamic display resolution (they increase or decrease screen resolution based on battery level I believe).

When using a persistent tile cache, this can cause mapsforge to load cached tiles that were generated when the phone was in a different resolution.

IMO, tile cache invalidation should happen automatically a tile size change is detected.

Example of what a user might see:

![Cache tiles being loaded](https://user-images.githubusercontent.com/1402931/75363612-a4c8b600-58ba-11ea-9b89-43700220efb4.jpg)

devemux86:
Not everyone will like such automatic actions and I/O operations should be avoided inside UI calls.

See older [TileSizeChanger](https://github.com/mapsforge/mapsforge/blob/2e4f936b9b017124b55a78d60ee84e52dbb8f58b/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/TileSizeChanger.java) sample where is demonstrated how to properly handle such situations.

sudokai:
Thanks for the feedback. I've created a forum topic: https://groups.google.com/forum/#!topic/mapsforge-dev/lMVMrgSBPdU

I'd appreciate if we could continue the discussion in the forum.

devemux86:
> I'd appreciate if we could continue the discussion in the forum.

Indeed please first should discuss in forum any ideas or questions for the library.

PR #1152, opened 2020-02-26
fix: cache invalidation doesn't happen if tile size changes

sudokai:
Purge the tile cache if tile size changes.

Some phones such as Huawei phones have dynamic screen resolution
and cached tiles generated via a different resolution should not
be reused.

Issue: ttps://github.com/mapsforge/mapsforge/issues/1151


devemux86:
See https://github.com/mapsforge/mapsforge/issues/1151#issuecomment-591526859 and forum [discussion](https://groups.google.com/forum/#!topic/mapsforge-dev/lMVMrgSBPdU).

PR #1172, opened 2020-10-04
Allow setting "position" attribute to symbol and lineSymbol rendering instructions

b3nn0:
Hi,
This PR allows setting the "position" attribute to the `<symbol..>` and `<lineSymbol...>` rendering instructions.
This allows placing a Symbol explicitly above/below/left/right/... of the respective point (or line-center).
Therefore, it makes it possible to use symbols that have a specific hotspot, so only make sense if they are placed accordingly (such as a marker pin, an arrow, ...)

This also deprecates the `align-center` attribute for lineSymbol (which was only used for oneway roads in the default themes), as it is now replaced by `position="center"`. However, I left it in there for backwards compatibility.
The changes should be 100% backwards compatible.

Note that the default for `<symbol...` is center, while the default for `<lineSymbol...` is bottom_right. 
Even though this is not 100% intuitive, IMO, this is to maintain backwards compatibility for the cablecar tags in the default themes (and any other theme out in the wild).

Example: I tested with the (enlarged) traffic light icon in the sample app.
Default center:
grafik
Right:
grafik
Above left:
grafik

You get the idea :)

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Position.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Position.java
@@ -45,4 +45,20 @@ public enum Position {
         throw new IllegalArgumentException("Invalid value for Position: " + value);
     }
 
+    public boolean isOnLeftSide() {
+        return this == LEFT || this == ABOVE_LEFT || this == BELOW_LEFT;
+    }
+
+    public boolean isOnRightSide() {
+        return this == RIGHT || this == ABOVE_RIGHT || this == BELOW_RIGHT;
+    }
+
+    public boolean isOnUpperSide() {
+        return this == ABOVE || this == ABOVE_LEFT || this == ABOVE_RIGHT;
+    }
+
+    public boolean isOnLowerSide() {
+        return this == BELOW || this == BELOW_LEFT || this == BELOW_RIGHT;
+    }
+
 }
--- a/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/SymbolContainer.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/SymbolContainer.java
@@ -21,28 +21,42 @@ import org.mapsforge.core.model.Point;
 import org.mapsforge.core.model.Rectangle;
 
 public class SymbolContainer extends MapElementContainer {
-    final boolean alignCenter;
     public Bitmap symbol;
     public final float theta;
+    public Position position;
 
-    public SymbolContainer(Point point, Display display, int priority, Bitmap symbol) {
-        this(point, display, priority, symbol, 0, true);
+    public SymbolContainer(Point point, Display display, int priority, Position position, Bitmap symbol) {
+        this(point, display, priority, position, symbol, 0);
     }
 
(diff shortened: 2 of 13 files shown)


b3nn0:
Sorry for the premature PR. Just committed another patch - now a Caption that references a Symbol will also respect the symbols position attribute when computing rendering offsets.
I.e. if the caption is positioned "below", it will be ensured to be below the symbol, no matter what the symbol's position is, etc.

Example: Bus Stop moved to above_right and related caption moved to above_right as well:
grafik

(for comparison, bus stop center would be here:)
grafik

devemux86:
_[Before you start working on an unresolved issue or try to implement a new feature](https://github.com/mapsforge/mapsforge/blob/master/docs/CONTRIBUTING.md), please contact us via our public [forum](https://groups.google.com/group/mapsforge-dev)._

_Also it's better to keep the pull requests simple, e.g. each theme rule can handled separately, instead of all mixed together._

On the topic, what is the actual use for all those changes?

Symbols are meant to be placed at the center of their position or can use pin symbols if want them to appear at bottom center.

Line symbols at different positions don't have much meaning either, since the lines have curves and the symbols will appear on all strange positions.

b3nn0:
Sorry for not using the forum first, didn't notice that.

The goal of this PR is to make it possible to actually use pin/marker-like symbols in a useful way.

For points:
E.g. an icon like this: https://www.onlinewebfonts.com/icon/548403 can then be used and it will actually point to the correct location if position="above" is configured - and even potentially show a label above or below the pin.
To make this possible in the current version, one would have to double the image size vertically, with the lower half being transparent, so that the pin's tip is actually in the center of the image - this is of course more of a workaround than a solution.
The same of course if the pin points to the bottom left - place it above_right, etc.

For lines:
this is - admittedly - less useful, but used in the default styles.
For oneway streets, there is align-center="true". For gondola, ski lifts etc, this is not given, which places the icons beside the line.
The reason I implemented this for lines as well is mainly because it gives more control over the rendering and for consistency.
E.g. you could place the same marker icon I linked above along a line and it will actually point/sit on the line instead of always being on the bottom-right of it.
Again, this change doesn't break anything. Default for lines is still below_right. And align-center=true is mapped to position="center".

About the traffic light example: The intention there was not to show traffic lights on the correct side of the road or something. It was just an example where it's easy to see what happens when setting position=..., as the actual coordinate is in the center of the crossroads.


Note that the actual changes to the code aren't even a lot. It is a fairly simple change.
The only reason why so many files are touched is, that the image boundary is now passed around instead of the actual image bitmap and the align-center value.
IMO, it also makes the code slightly simpler. E.g. the `computeHorizontalOffset` and `computeVerticalOffset` were in my opinion a bit hard to understand, and are now very straight forward.

devemux86:
Thanks for the extra details.

I merged it in branch [issue_1172](https://github.com/mapsforge/mapsforge/tree/issue_1172) with minor improvements,
so that everyone can test and make sure that library continues to work as before.

b3nn0:
Cool, thank you.
Already making use of it in my themes.

devemux86:
Thanks, merged via be09104f159e7aeadf451eb449e6bd46df59e44d.

PR #1173, opened 2020-10-06
fix ParallelPath problem with almost 180 degree angles

mg4gh:
As discussed in 
https://groups.google.com/forum/#!topic/mapsforge-dev/Gj9EtsdEpMI


devemux86:
Thanks, merged via 7115fdd84761d99094b442b57abb84d48abca312.

Issue #1176, opened 2020-10-21
Multithreaded rendering: race condition with Paint objects for area symbols


b3nn0:
Hi,
I found an issue with multithreading. For testing, I set `NUMBER_OF_THREADS = 20` in the AWT sample. In theory it can also happen with the default 2 threads, but much more unlikely.
Result: area symbol patterns look wrong - the pattern doesn't match from tile to tile:
![grafik](https://user-images.githubusercontent.com/1858945/96753243-da815880-13cf-11eb-9a68-03ad118ea74d.png)

The errors are random. Looks different after each restart.

Reason:
After several hours of debugging, I found the reason to be buried quite deeply:
In Area.renderWay, we set `fillPaint.setBitmapShaderShift()` to the correct offset that is required for drawing, then call `renderCallback.renderArea(...)`.
However, renderArea does not actually render, but only stores the paint and geometry.

However, this Paint object is recycled. When another thread tries to render a closed way, it will also reach this code location and set the same Paint object to this polygon as well (and modifies its BitmapShaderShift).

When the first thread now reaches the code that actually draws the geometry, it will now render with incorrect shift.

The only obvious solution that I see is, that the Paint object is actually copied for every instance. I.e. instead of
```
renderCallback.renderArea(renderContext, fillPaint, ..)
```
call
```
renderCallback.renderArea(renderContext, graphicsFactory.createPaint(fillPaint), ..)
```
and additionally fix the AwtPaint copy-constructor to actually copy the TexturePaint information as well (which it doesn't do right now).

The same issue might exist for other instructions as well. I have not checked yet.

- Is there any interest to make multithreaded rendering actually work?
- Any better idea for a fix?
If desired, I will create a pull request with my suggested fix.

devemux86:
_For questions or discussion please use the [forum](https://groups.google.com/group/mapsforge-dev)._

Creating objects continuously on each render call is usually not recommended for performance.
Multiple threads is an experimental feature - not for production and certainly not with 20 threads.

PR #1177, opened 2020-10-22
Fix for #1176: apply shader bitmap shift at drawing stage, not at processing stage

b3nn0:
Note that the setBitmapShaderShift method does indeed do some allocations, but this was done on a per-geometry basis previously as well (but then potentially overwritten by other threads), so nothing is lost here.


devemux86:
Thanks merged via ebe58bdef9c5a06e285ecd6fd5bdb4bac1d8baff, making dependent on threads number.

PR #1178, opened 2020-10-24
FrameBufferHA3: Improved version of FrameBufferHA2

bailuk:
**Add FrameBufferHA3 and make it default:** 
HA3 fixes an issue of HA2: FrameBufferHA2 sometimes prevents the LayerManager thread from terminating and then does not free its resources.

**Changes:**
- Reorganize class hierarchy of FrameBuffer classes.
- All FrameBuffer classes are now a subclass of abstract FrameBuffer
- Remove obsolete FrameBufferHA 
- FrameBufferOld (original MapsForge FrameBuffer) and FrameBufferH2 are unchanged and can still be used as a fallback
- Add improved FrameBufferHA3 and make it default. Can be configured by changing `FrameBufferController.FRAME_BUFFER_HA3`


devemux86:
Thanks for the clean work and the separate commits.
(makes easier the review)

Merged via 2679747d561253478d5aba09aad4310edbe86da3 (with the option in `Parameters` class).

PR #1181, opened 2020-11-29
Zipped hgt

usrusr:
Extended hillshading file reading to also support zipped .hgt, more specifically zips that follow the same tile naming convention as .hgt and contain an .hgt file with the appropriate name. All other .zips and all other contained files are ignored (e.g. no support for multiple tiles in one zip).

Rationale: distinguished tile librarian Sonny is offering downloads in the form of single-tile zips (which isn't very convenient when done manually) or a zip of single-tile zips. Manually unzipping the single tiles would be burdensome, waste disk space on uncompressed volumes. As an unintended side effect, in many cases (particularly on tiles that contain a lot of ocean) CPU for decryption might be faster than the extra I/O bandwidth required for an uncompressed read. If both versions exist, the one with the bigger resolution (internal file size) will win, first by filesystem iteration order on tie.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AbsShadingAlgorithmDefaults.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AbsShadingAlgorithmDefaults.java
@@ -1,0 +1,83 @@
+package org.mapsforge.map.layer.hills;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+public abstract class AbsShadingAlgorithmDefaults implements ShadingAlgorithm {
+    private static final Logger LOGGER = Logger.getLogger(org.mapsforge.map.layer.hills.AbsShadingAlgorithmDefaults.class.getName());
+
+    @Override
+    public RawShadingResult transformToByteBuffer(HgtCache.HgtFileInfo source, int padding) {
+        int axisLength = getAxisLenght(source);
+        int rowLen = axisLength + 1;
+        InputStream stream = null;
+        FileChannel channel = null;
+        try {
+            ByteBuffer map = null;
+            File file = source.getFile();
+            String nameLowerCase = file.getName().toLowerCase();
+            if (nameLowerCase.endsWith(".zip")) {
+
+                ZipInputStream zipInputStream = new ZipInputStream(new FileInputStream(file));
+                ZipEntry entry;
+                String expectedNameLC = nameLowerCase.substring(0, nameLowerCase.length() - 4) + ".hgt";
+                while (null != (entry = zipInputStream.getNextEntry())) {
+                    if (!expectedNameLC.equals(entry.getName().toLowerCase())) continue;
+
+                    map = ByteBuffer.allocate((int) entry.getSize());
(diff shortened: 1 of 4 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/ShadingAlgorithm.java
```diff
@@ -17,8 +17,86 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 
 public interface ShadingAlgorithm {
+    abstract class AbsShadingAlgorithmDefaults implements ShadingAlgorithm{
```
devemux86, line 32: Please separate `ShadingAlgorithm` from its nested classes (old and new) and use a separate file for each class for simplicity / reading reasons.

devemux86:
Is there the same code in the 2 `transformToByteBuffer`?
You can use a separate class for `AbsShadingAlgorithmDefaults` and avoid code duplication.

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/ShadingAlgorithm.java
usrusr: Done. That nested parent pattern is basically a crude backport of the Java 8 interface extensions (default methods etc) but I see how avoidable inheritance always comes with a readability cost.

Would raw DEM reading code deduplication implemented as a static method on RawShadingResult (or even as constructor) be OK?

devemux86: > That nested parent pattern is basically a crude backport of the Java 8 interface extensions

Then we should simplify completely the `ShadingAlgorithm`.
The library must support Java 7 for maximum compatibility.

usrusr: Yes, the transformToByteBuffer are duplicate. 

Separate class for AbsShadingAlgorithmDefaults is fine as well (convert(..) is only an internal sub-step of the public transformToByteBuffer so it would preferrable remain non-public so an abstract class is cleaner than some static helper method).

Hard-pushing an amended commit if fine for github and your preference for squashedness, right? (I'm more of a gitlab and messy unaltered commit history person, eager to see other approaches)

usrusr: > 
> 
> > That nested parent pattern is basically a crude backport of the Java 8 interface extensions
 
> The library must support Java 7 for maximum compatibility.

The pattern doesn't use Java 8 features, it tries to emulate as good as it can using nothing than Java 7

But with the top level AbsShadingAlgorithmDefaults it's conventional j7 style anyways

usrusr: Forced the manually squashed (rollback+amend) top level AbsShadingAlgorithmDefaults version.

Is github smart/aggressive enough to not pollute history with that Merge pull request commit?

devemux86: The pull request has 2 commits: a strange one and the correct with the code changes.

usrusr: Seems like my handling of upstream isn't is it should be. 

My zipped-hgt branch seems fine now, fork master required some more hand-holding (not that it matters for this PR)

devemux86:
Thanks, merged via 7ee8cfa049e763eadef04e116afff1a3f92796ee.

PR #1186, opened 2021-01-09
Xml Themes: support Android scoped storage and custom ressource providers

eddiemuc:
As agreed with @devemux86 in https://groups.google.com/g/mapsforge-dev/c/4LcJL93sa14:

This is a PR to introduce scoped storage support also for resource provisioning into mapsforge.

MapsForgeViewer is adapted as well to give usage examples.


devemux86:
The "Theme with resources (API21 req.)" menu does not seem to work?
Only an empty white map appears.

eddiemuc:
This should work...
What device (on which API level) are you using? I tried emulator with API 29, would like to retest on the level you are using.

Seen any exceptions in log?

devemux86:
I tried Android emulator with Android 11 (API 30).

eddiemuc:
OK, I tried too. Looks like it is working but is really really (far too) slow! I had to wait 10 seconds for the map to reappear.
Is it the same at your device?

I will investigate this further

devemux86:
> Looks like it is working but is really really (far too) slow! I had to wait 10 seconds for the map to reappear.

Indeed, if we wait long enough then the map is rendered.

eddiemuc:
I put some logs inside and measured time. This is crazy!

In my example I have about 400 icons to load. To get the 400 Uris the system needs about 1 second. For each of these files, to open an InputStream out of the Uri (not to read the actual data, just to open the Stream using contentResolver.openInputStream()) it takes about 20ms. 400 x 20ms = 10 seconds!

I don't currently know what to do about this. Have to think a bit about that.

eddiemuc:
Hi @devemux86, 
in context of the performance finding I had a new idea: I created a "ZippedXmlThemeResource" which is able to read xml themes and their resources out of a Zip file (e.g. like the ones provided by https://www.openandromaps.org/). 
When using this together with Android scoped storage, at least Android has only to read in ONE file (although it is a bigger one).

ZippedXmlThemeResource reads a provided Zip-Stream and keeps the uncompressed data (xml themes and resources) in memory. I pushed the stuff to this PR, example is contained in MapsforgeMapViewer.

What do you say

devemux86:
Nice idea, this seems a lot more convenient.
Have you finished with the changes?

eddiemuc:
Hi @devemux86 ,

Just pushed a stabilization fix for the zipped resource provider to this PR. Also I though how to best continue with this.

The PR currently contains three different things wrt Xml Theme Resource Provisioning
* **Base**: introduction of XmlThemeResourceProvider, its integration into StreamXmlTheme and its usage in XmlUtils. This is the basis for all other work, and I beg you to accept it in mapsforge
* **Zipped**: ZipXmlThemeResourceProvider as implementation of XmlThemeResourceProvider for providing resource in a zipped file. I think this is somethink useful for all mapsforge clients and imho should become part of mapsforge. If you disagree, that's also fine for me - as long as "base" is inside mapsforge I am able to implement this just for c:geo usage
* **Content**: ContentResolverResourceProvider as implementation of XmlThemeResourceProvider and used inside Content to provide resource files through Android scoped storage. I don't think we will use this in c:geo because it is simply too slow (I guess we will either only provide support for zipped ressources in the future or copy xml theme resources to an internal file dir before usage, haven't decided yet). Likewise I personally don't think this provides much value to mapsforge library. However, if you are still interested it can be integrated of course

What is your opinion aboiut all of this. Depending on it I can provide separate PRs for the different parts (**base**, **zipped**, **content**, always inculding the necessary parts in the Android test app) or fix up a single PR with just the parts you want to take over in mapsforge. My personal recommendation would be to take over **base** and **zipped**, but leave out **content**.  I really really beg you to allow **base** inside mapsforge - otherwise it gets very ugly in c:geo.

devemux86:
Thanks for the extra details.

The changes seems clean to we can continue with this pull request.

Please give me some time to review the code.

Since Mapsforge is developed along with [VTM](https://github.com/mapsforge/vtm),
I also have to check how all this applies there.

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/ZipXmlThemeResourceProvider.java
```diff
@@ -0,0 +1,102 @@
...
+                byte[] entry = readComplete(zipInputStream, (int)zipEntry.getSize());
+                String fileName = zipEntry.getName();
+                if (fileName.startsWith("/")) {
+                    fileName = fileName.substring(1);
+                }
+                files.put(fileName, entry);
+                if (zipEntry.getName().endsWith(".xml")) {
+                    xmlThemes.add(zipEntry.getName());
+                }
+            }
```
devemux86, line 60: We should check also for the theme archives that can contain a root folder inside the zip.

devemux86:
I merged the pull request in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186) with:
- simplifying the samples
- minor improvements
- a fix to read theme archives which contain a root folder

Please check and test that branch until I make more tests too.

eddiemuc:
> I merged the pull request in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186) with:
> - simplifying the samples
> - minor improvements
> - a fix to read theme archives which contain a root folder
> 
> Please check and test that branch until I make more tests too.

Hi @devemux86,

looking at the code it looks good :-)
I try to do a real test tomorrow and give you feedback.

Thanks for your great support
  Eddie

eddiemuc:
Hi @devemux86,
I checked branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186) with following result
* you removed my "blue star" example - which is fine, it was probably too "disturbing" anyway :-)
* you are right wrt to the "root folder" problem. But I think your fix will fail if the zip is NOT a root folder zip but by accident starts with a folder (e.g. "ele_res"). Also, you made me realize that the problem basically is that all file references are simply relative to the xml theme files and thus there can only be only one folder containing xml theme files.

For the second one I can provide a fix. Should I make a PR for this towards branch issue_1186?

devemux86:
> But I think your fix will fail if the zip is NOT a root folder zip but by accident starts with a folder (e.g. "ele_res").

Are you sure? With what theme zip we can check that?

It was tested and it works with original Elevate theme which has not everything under a root folder,
but has only directly the "ele_res" folder and xml files.

eddiemuc:
You can:
* take two different archives for two different themes (e.g. one from openandroidmaps and one from freizeitkarte)
* create some root folder (e.g. "mythemes"), extract both of the archives underneath it
* zip up the root folder again

I created PR #1187 against branch issue_1186. This one can even handle this scenario. It uses the already existing "relativePath" also for finding resources inside the zip

devemux86:
> I created PR #1187 against branch issue_1186.

Thanks, merged in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186).

But now the `ContentResolverResourceProvider` does not seem to work anymore?

eddiemuc:
> But now the `ContentResolverResourceProvider` does not seem to work anymore?

My apologies. PR #1189 (also against branch issue_1186) will fix this.

devemux86:
> PR #1189 (also against branch issue_1186) will fix this.

Thanks, squashed and merged in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186).

Let's test it more and see if it needs any improvements.

devemux86:
I pushed some more improvements,
like using a `File` instead of `XmlTheme` for a simpler implementation.

@eddiemuc please also check if everything works fine.

eddiemuc:
> I pushed some more improvements,
> like using a `File` instead of `XmlTheme` for a simpler implementation.
> 

Hm I don't think it is such a good idea to misuse the File class for this case. Because the values inside it do not represent a File, just an entry inside a Zip archive. Maybe just rename class XmlTheme to "ZipEntry" instead? Have you checked whether the File approach works well in scoped storage environment?

devemux86:
> Have you checked whether the File approach works well in scoped storage environment?

I have not seen any problems, please check that too.

devemux86:
> Because the values inside it do not represent a File, just an entry inside a Zip archive.

They are paths, we could use `String`, but want to avoid complicated implementation and not friendly user experience.

eddiemuc:
> > Because the values inside it do not represent a File, just an entry inside a Zip archive.
> 
> They are paths, we could use `String`, but want to avoid complicated implementation and not friendly user experience.

But `File ` is not a class representing an abstract Path. It represents a local File entity and has a rich interface which does not work as expected for these entries. E.g. calling ``exists()`, `isFile()` or `mkdirs()` on such an entry will definitely not trigger results that can be expected from a File object.

I imagine you did this because you don't like the overhead of creating a value class for this case, correct? If this is true, what about using something like `org.apache.commons.lang3.tuple.ImmutablePair` instead? Or maybe `java.net.Uri` (one can argue that this is a ressource identifier)?

devemux86:
> But File is not a class representing an abstract Path.

But it provides all functions which were implemented again in a new class.
i.e. find the xml theme file name and it's (relative) parent path (file functions)

> I imagine you did this because you don't like the overhead of creating a value class for this case, correct?

Exactly.

> what about using something like org.apache.commons.lang3.tuple.ImmutablePair instead?

We cannot add third-party dependencies in the library for such simple things.

> Or maybe java.net.Uri (one can argue that this is a ressource identifier)?

How will that be better or simpler to handle?

eddiemuc:
Hi @devemux86 
imho creating a small value class would be the best solution to the current situation. If you want to stay with File then lets go for it. But I imagine it will lead to many questions on user side. I advise to document explicitely that the `File `objects returned are not representing Files and can't be used according to the `File `class documentation.

Another example: if I have a `File`, I expect that I can read its content using `new FileInputStream(file)` or write to it using `new FileOutputStream()`. This will of course not work for the ones returned in this case.

Compared to returning a `File `I would personally prefer to just return a String with the full filename. Then it is up to the method's user to decide whether he wants to use a File object to parse it into path and name, which would be simple enough for him (just a one-liner).

devemux86:
This is work in progress, there is no rush to merge immediately, so we can think how it can be improved. :slightly_smiling_face:

eddiemuc:
Hi @devemux86 
I discovered a problem with the new Code: when run on windows, `File `internally uses `\` as separator while Zip always uses `/` (also on windows). This leads to failures, properly refleced by failing unit test `ZipXmlThemeResourceProviderTest.openZip` when executed on windows. Here is a screenshot in debug mode where you can see that the returned `File` paths contain `\` while the Zip-ressource uses `/` and thus a match fails.

image

My suggestion is to replace return value for `ZipXmlThemeResourceProvider.getXmlThemes()` from `List<File>` to `List<String>`. If you don't like this at all, what about using List<String[]> with two-element-array (first containing path, second filename)?  


It might be noted that the user (if he wants to and windows is not a supported system for his app) can extract path and filename from the string using easy twoliner e.g. like:
String name = new File(xmlThemePath).getName();
String path = new File(xmlThemePath).getParent().getPath()

Other users may decide to use other options depending on their needs, supported operating systems, already existing or included Utils-functions or other criterias. This should imho not be the decision of mapsforge.

devemux86:
Thanks for the testing.

> My suggestion is to replace return value for ZipXmlThemeResourceProvider.getXmlThemes() from List<File> to List<String>.

That seems the simplest solution.

I also added a `ZipRenderTheme` (optional) class, so that users can more easily use a zipped theme.

eddiemuc:
Hi @devemux86 ,

I like your solution very much, especially the introduction of `ZipRenderTheme`. 

I especially like your use of the codeline `xmlTheme.substring(0, xmlTheme.lastIndexOf("/") + 1)` which is a very clever one-liner working on both the case with and without subdirectories!

I did all my tests (unit + mapsforgeviewer with lots of different archive files as well as the contentstorage-scenario), from my point of view we are good to go. Thanks a lot for your help and work!

eddiemuc:
Hi @devemux86 , 

since the change seems to be stable and lying in branch issue_1186 now I guess I can close this PR?

For further planning in c:geo: what do you think when this change might go into master as well as a next mapsforge release (we can use it only after this happened of course)?

devemux86:
> since the change seems to be stable and lying in branch issue_1186 now I guess I can close this PR?

We can keep the PR open, as the discussion and possible improvements happen here.

> what do you think when this change might go into master as well as a next mapsforge release

I just made a [release](https://github.com/mapsforge/mapsforge/blob/master/docs/Changelog.md) some days ago.
We must test carefully any code changes first with master and then I can make a release in the future.

You can use the latest Mapsforge via its [snapshot builds](https://github.com/mapsforge/mapsforge/blob/master/docs/Integration.md#snapshots).

We also need to test all render theme features with the PR, like the [style menu](https://github.com/mapsforge/mapsforge/blob/master/docs/Rendertheme.md#stylemenus), etc.

devemux86:
I updated all themes so that they have a more simple and clean implementation.

devemux86:
Merged via cdccd83a18a8cb424cab28b3ef4a799ad369ca52.

Thanks again for the clean work and the useful contribution and testing!

eddiemuc:
> Merged via cdccd83a18a8cb424cab28b3ef4a799ad369ca52.
> 
> Thanks again for the clean work and the useful contribution and testing!

You are very welcome, and many thanks back for your support and your patience.

PR #1187, opened 2021-01-11
rel to #1186: allow handling of relative xml themes in zip

eddiemuc:
rel to #1186: allow handling of relative xml themes in zip


devemux86:
Thanks, merged in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186).

PR #1189, opened 2021-01-12
fix contentprovider

eddiemuc:
fix contentprovider


devemux86:
Thanks, squashed and merged in branch [issue_1186](https://github.com/mapsforge/mapsforge/tree/issue_1186).

PR #1190, opened 2021-02-21
ZipXmlThemeResourceProvider: add method to just scan zip for xmlthemes

eddiemuc:
We have the use case that when we have multiple Zip files in a directory, each containing XmlThemes, then we want to offer the user a selection for one of those XmlThemes. 

**Current situation:**
In order to get a complete list of all available XmlThemes we have to create an instance of `ZipXmlThemeResourceProvider ` for each Zip file, then to use the `getXmlThemes `method on them, then release this instance again.

However, this reads all of those zip files content into memory (including e.g. all image contents) and just throws this away afterwards. This takes long time and does not use memory well.

**Improvement:**
This PR adds a static method `ZipXmlThemeResourceProvider.getXmlThemesIn` which can be used to just scan a Zip file for contained XML Theme files without reading the whole Zip and contained resources into memory


devemux86:
Thanks, nice improvement.

Merged via b8a95f2b04878b646f70e37cc598373195978c9c, also updating the samples.

PR #1191, opened 2021-02-28
ContentResolverResourceProvider should handle document Uris correctly

eddiemuc:
ContentResolverResourceProvider should handle document Uris correctly

While using ContentResolverResourceProvider inside c:geo, I realized that it is currently unable to handle Uris as roiots which are NOT directly taken from ACTION_DOCUMENT_TREE intent but subdirectories of it (it converts all given Uris back to an Uri pointing to the root directory as selected by User). To allow this it must be able to handle already converted document Uris as well. I introduced a flag for that and a longer comment to explain it. Default behaviour is preserved

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/ContentResolverResourceProvider.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/rendertheme/ContentResolverResourceProvider.java
@@ -40,6 +40,7 @@ public class ContentResolverResourceProvider implements XmlThemeResourceProvider
 
     private final ContentResolver contentResolver;
     private final Uri relativeRootUri;
+    private final boolean isDocumentUri;
 
     private final Map<String, Uri> resourceUriCache = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


devemux86:
Thanks, merged via 9cdb3396fadc86358c0a08c69d42edd051b6e216.

devemux86:
@eddiemuc 
[DocumentsContract](https://developer.android.com/reference/android/provider/DocumentsContract) has several `isX` methods.
Can't we use one of these checks and avoid the extra parameter?

eddiemuc:
@devemux86 I thought about this too, but couldn't find one really serving this purpose:
* there is `isDocumentUri `which probably would be the right one to use, but is has a context parameter which I would have to introduce into `ContentResolverResourceProvider `just for this check (and also I don't know if the "additional" checks are restricting this too far)
* there is `isTreeUri`, but it turns out that a document uri can also be a tree uri at the same time so the check is not what we want (also, the method is only available starting with API level 24 /Android7.0)

So I decided to make the "pure" Uri pass if the user wishes (then the caller has full control to convert the Uri in case I oversee some special case where this might be necessary) and just handle the special but probably very common case of passing an Uri directly from an `OPEN_DOCUMENT_TREE `intent. Overall it seemed the cleanest way to me, and also open enough for users to bypass any unforeseen obstacles.

If you wish, we can introduce the `isDocumentUri` check instead, but then we have to pass a `Context `object to the class (instead of a `ContentResolver`, this resolver could be then extracted from the `Context`). Also I can't guarantee you now that the isDocumentUri check will always be what we want.

devemux86:
Thanks for the explanation, we can leave it as it is now, it seems the simplest implementation without `Context` dependency.

PR #1234, opened 2021-08-10
ConsumableEvent for InputListener

TheDoctorOne:
Added alternative way to check Circle.contains: "Circle.contains(MapView, Point)".
Implemented Consumable Event for onMove and onZoom events.

I needed these functions while working with 'java.awt', figured people also may need them.


devemux86:
Thanks for the suggestions, but these seem more intended for custom app needs.

- Better to extend / override the library API for custom app needs, rather than loading the library with custom code.
  (and objects that are constantly created with each gesture / event, reducing library performance for all users)
- It is essential that each pull request must implement only 1 feature.
- And everything must first be discussed and agreed in Discussions.

Let's not complicate things and keep the map library and custom app implementations separate.

---

_For questions, reports and feature requests please use the Discussions._

PR #1256, opened 2021-10-22
fix potential ArrayIndexOutOfBoundException in TouchGestureHandler

eddiemuc:
fix potential ArrayIndexOutOfBoundException in TouchGestureHandler

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/input/TouchGestureHandler.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/input/TouchGestureHandler.java
@@ -31,6 +31,7 @@ import org.mapsforge.core.model.Point;
 import org.mapsforge.core.util.Parameters;
 import org.mapsforge.map.android.view.MapView;
 import org.mapsforge.map.layer.Layer;
+import org.mapsforge.map.layer.Layers;
 import org.mapsforge.map.model.IMapViewPosition;
 
 /**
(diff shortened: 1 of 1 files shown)


eddiemuc:
Hi @devemux86 : a week has passed, and I have collected some results from Google Play console for cgeo beta containing the fix.

For explanation: the fix candidate is active in our beta version as of 24.10. This maps to apk 20211023.apk. The other beta versions (and productive versions) don't contain the fix. Here are all beta versions released in October 2021:
image
image

Here is the statistic for the ArrayIndexOutOfBoundsException in October 2021 (by app version):
image

The picture shows us:
* The error is very relevant to c:geo with >1000 occurences (each meaning an app crash) in October 2021 alone spread over many different app versions 
* Beta with fix is active now for 7 days without a single occurence of the ArrayOutOfBoundsException
* In constrast in previous two betas (running since 11.10. / 21.10.) the bug occured 53 times 

To interpret these results, one needs to know that (naturally) our user base for betas is significantly smaller than production. Also, the app c:geo is usually used much more intensive on weekends than on workdays (where many people need to work and have no time for geocaching) - so it is important to include "weekend data" into the statistics. I will continue to keep an eye on this.

In summary, I think this shows that the fix is working and it is relevant. What do you think?

devemux86:
Thanks for the feedback, these seem to be nice results.

Do you plan to include the change in release (as mentioned in https://github.com/cgeo/cgeo/issues/10128) for more time
to see extended test results from your entire user base?

More users and time (like 2 weekends) would help to see if the problem is solved this way.

eddiemuc:
Hi @devemux86 , I assume we will include the fix in next release, providing nothing unexpected comes up during remaining beta. Current beta is based on a modified fork of mapsforge on my own repository, I will probably move the used version to a fork on cgeo for better consistency. 

However, I don't want to make cgeo use a forked, self-modified mapsforge version for too long since this cuts us out from future releases. So at some point I would be glad if this could go to mapsforge master (except of course unexpected problems show up during ongoing evaluation).

eddiemuc:
Hi @devemux86, just a quick update on this. The error has not occured in beta ever since we used the installed fix, no other problems have occured wrt mapsforge. Beta has not been released to production yet (due to other, unrelated bugs and feature discussions) but the user base is already very large. 

I personally don't need more convincing that the PR solves the error but I can continue monitoring this of course. What would be the tipping point / proof needed from your side to merge this fix into mapsforge master?

devemux86:
Thanks for the work and the tests!

Merged via 5b81485f9a0d0e82a567c8c3bd483dae71f2806e.

eddiemuc:
Thanks a lot @devemux86 also for your support!

PR #1263, opened 2021-11-21
Fixes svg cache hashing algo when working with resource providers

eddiemuc:
Fixes svg cache hashing algo when working with resource providers

We discovered a problem in c:geo with the caching of svg images from XML render themes when using resource providers. Reason is that the hash calculated for the cache file relies on hashCode of the resource provider, but the resource providers don't implement this. Thus the hashCode is different on each app startup and hash value is different too on each startup -> rendering the caching useless across app restarts.

A detailled description of the resulting problem can be found here: https://github.com/cgeo/cgeo/issues/12145 

This PR will change this by using resource provider class name instead of hashCode().

@devemux86 : it seems I introduced this bug myself back when I created the resource provider things. Sorry for that.


devemux86:
Thanks for the fix!

PR #1266, opened 2021-11-24
Using TCK Tested JDK builds of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

To see more details of GitHub Action setup-java@v2 go to Foojay.io below:
https://foojay.io/today/github-actions-with-java-part-2/

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -17,7 +17,7 @@ jobs:
       uses: actions/setup-java@v2
       with:
         java-version: '11'
-        distribution: 'adopt'
+        distribution: 'zulu'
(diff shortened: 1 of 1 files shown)


devemux86:
Thanks for the suggestion, we use the official GitHub Action workflow.

Have you proposed GitHub to change the default JDK?
(then all projects will be more comfortable to adopt it)

---

_For questions, reports and feature requests please use the Discussions._

PR #1280, opened 2022-01-20
POI - Deal properly with equal sign in tag values

Juanjo-MC:
stringToTags method was silently ignoring tags whose value contains equal signs.

However, such values are legit in OSM. They may appear ie in urls containing query strings.

This PR fix that behaviour.

Discussion: https://github.com/mapsforge/mapsforge/discussions/1279


devemux86:
Thanks, merged via e2342b8a757684fb00574545f749afcec63dd4ad.

PR #1338, opened 2022-07-18
Android Storage Access Framework for hillshading (#1269)

usrusr:
Finally found some motivation to address #1269 

The regular java.io.File implementation is still in main (not awt), but actually that shouldn't be necessary given the minsdk 21? Android demo currently supports both

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
@@ -1,0 +1,52 @@
+package org.mapsforge.map.android.hills;
+
+import android.content.ContentResolver;
+import org.mapsforge.core.util.IOUtils;
+import org.mapsforge.map.layer.hills.DemFile;
+import org.mapsforge.map.layer.hills.DemFileFS;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public class DemFileAndroidContent implements DemFile {
+    private final DemFolderAndroidContent.Entry entry;
+    private final ContentResolver contentResolver;
+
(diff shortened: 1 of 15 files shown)


devemux86:
Thanks, merged via 43ade4a06a4c90dc8789656eed07a4995de0138c.

> The regular java.io.File implementation is still in main (not awt), but actually that shouldn't be necessary given the minsdk 21? Android demo currently supports both

Mapsforge is not only for Android, but should also work on Desktop.

usrusr:
I know: I wasn't talking about removal, only about moving the parts that wouldn't work on to the desktop-specific module. But when I wrote about perhaps moving it I wasn't considering that java.io.File isn't going anywhere on android for the private file areas, moving to mapsforge-map-awt would have been wrong.

Actually I don't even work on a mapsforge client for android, only on the desktop. [Quäldich Tourenplaner](https://tourenplaner.quaeldich.de/), an ancient desktop dinosaur where the inclusion of mapsforge feels like time-travel from the future. But I'm an android dev by day so leaving hillshading without SAF support felt wrong even without having a use case for the code. 

Thanks for the work you do on mapsforge! (and VTM!)

devemux86:
> java.io.File isn't going anywhere on android for the private file areas

Correctly, if the dem files are accessed in app storage, then it is easier / faster to use file api instead of streams.

---

By the way if you have free time, it would be nice to update the [HillshadingMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/HillshadingMapViewer.java) sample.
Use SAF to open a map before the dem, like the [GettingStarted](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/GettingStarted.java) and [MapsforgeMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MapsforgeMapViewer.java) samples.
(and not extend other activity)

PR #1356, opened 2022-09-01
#1355 use "tile.openstreetmap.org" as tile server as requested by OSM project

### LINKED ISSUE #1355: Use preferred tile.openstreetmap.org URL
> Many projects, examples, and documentation still use old URLs to access the `tile.openstreetmap.org` service. The URLs should be updated as follows:
> 
> * `a.tile.openstreetmap.org` -> `tile.openstreetmap.org`
> * `b.tile.openstreetmap.org` -> `tile.openstreetmap.org`
> * `c.tile.openstreetmap.org` -> `tile.openstreetmap.org`
> * `a.tile.osm.org` -> `tile.openstreetmap.org`
> * `b.tile.osm.org` -> `tile.openstreetmap.org`
> * `c.tile.osm.org` -> `tile.openstreetmap.org`
> * `tile.osm.org` -> `tile.openstreetmap.org`
> * `http://*` -> `https://` (use secure URL)
> 
> `tile.openstreetmap.org` supports [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) and [HTTP/3](https://en.wikipedia.org/wiki/HTTP/3) which no longer require the old (a|b|c) aliases to increase browser connection concurrency. Using a single URL improves performance and ability to cache.
> 
> The old URLs have a maintenance burden and add complexity. The old URLs may stop working in the future. Please switch to the preferred URL now.
> 
> (This issue is for linking 3rd party projects on Github and elsewhere which are being asked to update `tile.openstreetmap.org` URLs)

https://github.com/openstreetmap/operations/issues/737

ztNFny:
OSM requests that apps stop using the tileservers "a/b/c.tile.openstreetmap.org" and instead use "tile.openstreetmap.org" only
https://github.com/openstreetmap/operations/issues/737 

fixes #1355

### DIFF
--- a/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/DownloadCustomLayerViewer.java
+++ b/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/DownloadCustomLayerViewer.java
@@ -24,9 +24,7 @@ import org.mapsforge.map.layer.download.tilesource.OnlineTileSource;
 public class DownloadCustomLayerViewer extends DownloadLayerViewer {
     @Override
     protected void createLayers() {
-        OnlineTileSource onlineTileSource = new OnlineTileSource(new String[]{
-                "a.tile.openstreetmap.fr", "b.tile.openstreetmap.fr", "c.tile.openstreetmap.fr"},
-                443);
+        OnlineTileSource onlineTileSource = new OnlineTileSource(new String[]{"tile.openstreetmap.fr"}, 443);
         onlineTileSource.setName("Humanitarian").setAlpha(false)
                 .setBaseUrl("/hot/")
                 .setParallelRequestsLimit(8).setProtocol("https").setTileSize(256)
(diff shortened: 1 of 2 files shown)


mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/DownloadCustomLayerViewer.java
```diff
@@ -24,9 +24,7 @@
 public class DownloadCustomLayerViewer extends DownloadLayerViewer {
     @Override
     protected void createLayers() {
-        OnlineTileSource onlineTileSource = new OnlineTileSource(new String[]{
-                "a.tile.openstreetmap.fr", "b.tile.openstreetmap.fr", "c.tile.openstreetmap.fr"},
-                443);
+        OnlineTileSource onlineTileSource = new OnlineTileSource(new String[]{"tile.openstreetmap.fr"}, 443);
```
devemux86, line 27: This was not requested and is not working, have you tested it?

devemux86:
Thanks, merged the requested and working change via 08d1b6dd8adf5719f8e7427777d54bf2a30d8233.

mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/DownloadCustomLayerViewer.java
ztNFny: In https://github.com/openstreetmap/operations/issues/737 it was requested to use the tile servers listed in the wiki only.

Didn't test it, I trusted the official documentation @ https://wiki.openstreetmap.org/wiki/Tile_servers to be correct - which it isn't. So "use the official sources" has to be taken with a grain of salt here :-(

devemux86:
The tests are fixed via 0531de9b38ba10302625da9e6b863bab5532d4c2.

Issue #353, opened 2014-11-06
Render-theme with "mtb:scale:uphill" and "mtb:scale" displays wrong tags


mapsforge:
_From [frank.hu...@gmail.com](https://code.google.com/u/109115453950959521277/) on September 18, 2012 21:58:02_

What steps will reproduce the problem? 1. Create a small xml (osm) file rttesttrack.xml, modify some tracks and add mtb:scale and mtb:scale:uphill tags. Tags should be different (e.g mtb:scale = 1, mtb:scale:uphill=2)
1. Add the following lines after line 386 to the original tag-mapping.xml and save it as mtbtag-mapping.xml:
   
   ```
   <osm-tag key="mtb:scale" value="0" renderable="true" />
   <osm-tag key="mtb:scale" value="1" renderable="true" />
   <osm-tag key="mtb:scale" value="2" renderable="true" />
   <osm-tag key="mtb:scale" value="3" renderable="true" />
   <osm-tag key="mtb:scale" value="4" renderable="true" />
   
   <osm-tag key="mtb:scale:uphill" value="0" renderable="true" />
   <osm-tag key="mtb:scale:uphill" value="1" renderable="true" />
   <osm-tag key="mtb:scale:uphill" value="2" renderable="true" />
   <osm-tag key="mtb:scale:uphill" value="3" renderable="true" />
   <osm-tag key="mtb:scale:uphill" value="4" renderable="true" />
   ```
2. Convert the modified osm file with osmosis/mapwriter: 
   osmosis-0.40.1/bin/osmosis --rx file=rttesttrack.xml --mw file=rttesttrack.map
   bbox=48.48,9.17,48.5,9.19 tag-conf-file=mtbtag-mapping.xml
3. Modify an existing render theme (I took the bluenight-theme from Locus forum), add the lines to show up symbols for MTB (for those interested, I tried to make the same symbols as in openmtbmap.org map for Garmin devices). I will not show all tags, but the idea is: 

<rule e="way" k="highway" v="track|path">
  <rule e="way" k="mtb:scale" v="0" zoom-min="16">
    <lineSymbol src="file:/symbols/mtbscale0.png"
        align-center="false" repeat="true" />
  </rule>
  <rule e="way" k="mtb:scale" v="1" zoom-min="16">
    <lineSymbol src="file:/symbols/mtbscale1.png"
        align-center="false" repeat="true" />
  </rule>

[...]

PR #1366, opened 2022-11-24
Fix of problem with incorrect handling of nested tags

menion:
## Specification of the problem

Imagine following line
```
<?xml version='1.0' encoding='UTF-8'?>
<osm version='0.6' generator='JOSM'>
  <node id='-101755' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1' lat='50.01334453162' lon='15.75031220376' />
  <node id='-101756' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1' lat='50.01342037049' lon='15.7506501621' />
  <node id='-101757' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1' lat='50.01357549509' lon='15.75128316343' />
  <node id='-101761' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1' lat='50.01385816529' lon='15.75433015287' />
  <node id='-101762' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1' lat='50.0148957575' lon='15.75657247961' />
  <way id='-101802' action='modify' timestamp='2022-06-28T11:25:07Z' visible='true' version='1'>
    <nd ref='-101755' />
    <nd ref='-101756' />
    <nd ref='-101757' />
    <nd ref='-101761' />
    <nd ref='-101762' />
    <tag k='network' v='lwn' />
    <tag k='osm' v='yes' />
    <tag k='osmc_background' v='white' />
    <tag k='osmc_color' v='green' />
    <tag k='osmc_foreground' v='green_backslash' />
    <tag k='route' v='hiking' />
    <tag k='type' v='route' />
  </way>
</osm>
```

and a simple map theme
```
<?xml version="1.0" encoding="utf-8"?>
<rendertheme map-background="#ebeade" version="4" xmlns="http://mapsforge.org/renderTheme" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://mapsforge.org/renderTheme https://raw.githubusercontent.com/mapsforge/mapsforge/dev/resources/renderTheme.xsd">

	<rule e="any" k="*" v="*">
		<rule e="way" k="osmc_color" v="*" zoom-max="18" zoom-min="15">
			<rule e="way" k="osmc_color" v="green">
				<line dy="-0.9" stroke="#267f00" stroke-linecap="butt" stroke-width="0.6"/>				
			</rule>
			<rule e="way" k="osmc_color" v="white">
				<line dy="0.9" stroke="#EA3200" stroke-dasharray="8,6" stroke-linecap="butt" stroke-width="0.6"/>
			</rule>		
		</rule>
	</rule>
		
</rendertheme>
```

The result is following (incorrect)
image

### The reasons are two:

1. `PositiveRule` matches the` keys` and `values` of every tag separately (other rules as well of course)! Because the line contains `osmc_background = white` and `osmc_color = green`, this cases that both rules are applied. But there is no `osmc_color = white`!
2. Even after fixing problem 1, there remains an issue caused by the optimization. Optimization change third rule key from `osmc_color` to `AnyMatcher`and ... 

## Solution
The solution to this problem is in this PR.

1. All matches are made per `Tag` and not by all tags at once
2. Completely disabled optimization in the `RuleBuilder` class

The result is (correct)
image

## Benchmark
I use my own simple benchmark for testing of the tile rendering system.
Setup: OpenAndroMap, Elements theme, zoom level 12, 3000 tiles

Old solution: 106ms per tile
New solution: 105ms per tile

The test is simple and not perfectly precise, but at least it shows that removing of the optimization does not have a negative effect.

### DIFF
--- a/mapsforge-map-awt/src/test/java/org/mapsforge/map/rendertheme/rule/KeyMatcherTest.java
+++ b/mapsforge-map-awt/src/test/java/org/mapsforge/map/rendertheme/rule/KeyMatcherTest.java
@@ -42,7 +42,7 @@ public class KeyMatcherTest {
         Tag tag2 = new Tag(KEY2, null);
         AttributeMatcher attributeMatcher = new KeyMatcher(Arrays.asList(KEY1));
 
-        Assert.assertTrue(attributeMatcher.matches(Arrays.asList(tag1)));
-        Assert.assertFalse(attributeMatcher.matches(Arrays.asList(tag2)));
+        Assert.assertTrue(attributeMatcher.matches(tag1));
+        Assert.assertFalse(attributeMatcher.matches(tag2));
     }
 }
--- a/mapsforge-map-awt/src/test/java/org/mapsforge/map/rendertheme/rule/ValueMatcherTest.java
+++ b/mapsforge-map-awt/src/test/java/org/mapsforge/map/rendertheme/rule/ValueMatcherTest.java
@@ -42,7 +42,7 @@ public class ValueMatcherTest {
         Tag tag2 = new Tag(null, VALUE2);
         AttributeMatcher attributeMatcher = new ValueMatcher(Arrays.asList(VALUE1));
 
-        Assert.assertTrue(attributeMatcher.matches(Arrays.asList(tag1)));
-        Assert.assertFalse(attributeMatcher.matches(Arrays.asList(tag2)));
+        Assert.assertTrue(attributeMatcher.matches(tag1));
+        Assert.assertFalse(attributeMatcher.matches(tag2));
     }
 }
--- a/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/AnyMatcher.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/AnyMatcher.java
@@ -16,9 +16,8 @@ package org.mapsforge.map.rendertheme.rule;
 
 import org.mapsforge.core.model.Tag;
 
-import java.util.List;
-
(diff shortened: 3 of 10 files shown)


devemux86:
@menion thanks for the extremely useful contribution!

It appears to be about the infamous (and old) issue #353 about tags with same values.
([old forum discussion](https://groups.google.com/g/mapsforge-dev/c/FP-_bdmGJnQ/m/tpsaUBBlCQAJ))

Can test with tag-mapping:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<tag-mapping xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" default-zoom-appear="16"
    profile-name="default-profile" xmlns="http://mapsforge.org/tag-mapping"
    xsi:schemaLocation="http://mapsforge.org/tag-mapping https://raw.githubusercontent.com/mapsforge/mapsforge/master/resources/tag-mapping.xsd">

    <ways>
        <osm-tag key="network" value="lwn" zoom-appear="0" />
        <osm-tag key="osmc_color" renderable="false" value="green" />
        <osm-tag key="osmc_background" renderable="false" value="white" />
    </ways>

</tag-mapping>
```

And command line:
```shell
osmosis --rx file=test.osm --mw file=test.map tag-conf-file=tag-mapping.xml bbox=50.01334453162,15.75031220376,50.0148957575,15.75657247961
```

([VTM](https://github.com/mapsforge/vtm) does not seem to have such problem)

menion:
Hello Emux,
I've not tried VTM to be true, but expect an identical issue there, right? 
Glad you consider the solution as working and merged this PR. Thanks.

Issue #1085, opened 2018-09-08
Being able to render tile by tile


monsieurtanuki:
Regarding `DatabaseRenderer`:
* the purpose is to render tiles
* in the context of a visible map
* being called by an instance of `MapWorkerPool`

But that doesn't match a slightly different way of seeing things: a simple "give me that tile".
Because `DatabaseRenderer`:
* interacts with `MapWorkerPool` so that it forces a display refresh (`MapWorkerPool.this.layer.requestRedraw();`), that we don't want here
* includes some optimizations, like not writing twice the same label (cf. `Set<MapElementContainer> undrawableElements`), that provokes side-effects (cf. https://github.com/osmdroid/osmdroid/issues/1145)
* in this last case, potentially because of a bug - as elements are never removed from `tileDependencies` _except_ in `MapWorkerPool`, that we don't want to use because of the display refresh

My suggestion is to create a dedicated class that extends `StandardRenderer`.
The purpose is to have the same level of quality as `DatabaseRenderer`, but in a simpler context where we will avoid side-effects.
The name will be something like `DirectRenderer`.

devemux86:
Mapsforge is not meant to be a raster tile generator, there are specific tools for that, like the ones used by all online raster tile providers.

Also Mapsforge offers more advanced real-time rendering of labels + symbols in separate layer, which cannot be written in the raster tiles of the map.

BTW Mapsforge has not any rendering problems via its engine. So cannot know how any external implementations use its low level API correctly or not.

However if someone wants to provide pull request which:
- Doesn't break library's current working behavior
- Can work as extension to help such workflows

We can certainly review it.

devemux86:
There is the [SaveTiles](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-awt/src/main/java/org/mapsforge/samples/awt/SaveTiles.java) sample demonstrating similar workflow.

monsieurtanuki:
@devemux86 Oh of course I didn't mean to imply that rasters are better than vectors or that the only point of Mapsforge is to build rasters. Different technology, different usages.

Thanks for your SaveTiles sample. It's more or less what we used to do, and we may encounter the side-effects I described: if we repeatedly ask for the same neighboring tiles, sometimes the labels are not consistent anymore.

Btw I don't seem able to push, I get an error 403. Any idea why? (I'm not a git expert)

devemux86:
[...]

PR #1367, opened 2022-11-28
Tile rendering performance improvements

menion:
I would like to introduce three points that increase the performance of tile rendering. Every point is in a separate commit so it is visible what exactly has changed.

### 1. Improve iterations over tags
Rendering time reduced to 91%

In the old code, iterations were done as `for (Tag tag : tags)` call. This anyway creates an iterator on the background. Direct iteration over `for (int i = 0; ...)`, in the number of calls, cause a significant time reduction.

### 2. Convert `List<Tag>` to `Tag[]`
Rendering time reduced to 96%

We probably do not have to talk that `List` adds a little more work memory requirement than Array. Again, the difference is not big, but in the amount of points/ways and its tags, this makes a difference of 4% in speed! (and probably also save a GC a little bit).

### 3. Introducing keys & values in Tags as `int` values
Rendering time reduced to 91%

Comparing and testing Strings is definitely a lot slower than testing `int` values. Because of this, every `Tag` has now `keyCode` and `valueCode` values that are used for tests and matching.

Here is anyway something to discuss. In the Locus Map, for 5+ years I use simple `String.hashcode` to get the "unique" `int` value. This works without any known problem! But, `hashcode` may not return a unique value for 100%. I do not know any better generator of unique int/long values and because I do not think, the current solution is a problem, suggest using it. But just FYI.

## Summary
These three commits introduce three smaller (but tested by years) changes, that all together, reduce the rendering time of tiles to 79% of the original time. Suggest giving this branch a try and testing it personally, even in VTM. The difference is directly visible during map movement.

And feedback is, of course, welcome and I'll gladly improve it in case, the code does not match any "standards".


tqu-dev:
Are you sure that counters are faster than iterators? My tests show no difference (after a warm-up period).

menion:
I'm not a java-bytecode expert, so no, I'm not sure. On the second side, I wrote my own simple benchmark where the benchmark computes the time from the rendering start till all tiles on the screen are fully rendered. Then clear a cache and repeat it. This 100x.

I did these tests 3x after every change to see its differences and the results are published here. I made a few more changes, but the differences were not significant.

devemux86:
Thanks @menion, I will certainly check the pull request.

> even in VTM

VTM is compatible with Mapsforge maps + styles, but does not use these libs.
It has its own reader and repository: https://github.com/mapsforge/vtm

---

May I ask what Mapsforge you are using now on Locus?
I remember you used a modified v0.3, are you now using the snapshot from repository?

Do you use the [Label Layer](https://github.com/mapsforge/mapsforge/blob/master/docs/LabelLayer.md) or have implemented map rotation or fractional zoom?

devemux86:
> Instead of playing on my own playground, I would like to contribute to the official repo.

:+1

> currently use DirectRenderer as the source for map tiles

You probably mean https://github.com/mapsforge/mapsforge/issues/1085?

Rendering tile by tile may miss the Mapsforge cache optimizations for labels / symbols that overlap neighbor tiles.
See https://github.com/mapsforge/mapsforge/issues/1085#issuecomment-420086968 by @ludwigb

devemux86:
>  Improve iterations over tags

@menion have you checked if it is better:
```java
 for (int i = 0, n = tags.size(); i < n; i++) {
```
instead of:
```java
for (int i = 0; i < tags.size(); i++)
```

menion:
> > currently use DirectRenderer as the source for map tiles
> 
> You probably mean #1085?
> 
> Rendering tile by tile may miss the Mapsforge cache optimizations for labels / symbols that overlap neighbor tiles. See the [#1085 (comment)](https://github.com/mapsforge/mapsforge/issues/1085#issuecomment-420086968) by @ludwigb and this old [forum discussion](https://groups.google.com/g/mapsforge-dev/c/UA3EHDx9yDw).

Yes, this is correct. `DirectRenderer` has the option to register the so-called `TileRefresher` , which [triggers requests](https://github.com/mapsforge/mapsforge/blob/3f171f2a5076d746497f49a8587765034f3583b9/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DirectRenderer.java#L178) to reload a certain tile when there is any overlapping text. This seems to work correctly.

> > Improve iterations over tags
> 
> @menion have you checked if it is better:
> 
> ```java
>  for (int i = 0, n = tags.size(); i < n; i++)
> ```
> 
> instead of:
> 
> ```java
> for (int i = 0; i < tags.size(); i++)
> ```

This for sure will be a little faster. But I was not testing it directly, because the next step (step 2.) was conversion to arrays, where the test is made over the internal variable `length` that may be accessed directly without any slowdown.

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/StandardRenderer.java
```diff
@@ -186,7 +188,9 @@ protected void renderWaterBackground(final RenderContext renderContext) {
         for (int i = 0; i < coordinates.length; i++) {
             coordinates[i] = coordinates[i].offset(tileOrigin.x, tileOrigin.y);
         }
-        PolylineContainer way = new PolylineContainer(coordinates, renderContext.rendererJob.tile, renderContext.rendererJob.tile, Collections.singletonList(TAG_NATURAL_WATER));
+        Tag[] tags = new Tag[1];
+        tags[0] = TAG_NATURAL_WATER;
```
devemux86, line 192: Since this is a static array, we can create it once together with `TAG_NATURAL_WATER`?

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
```diff
@@ -184,6 +188,9 @@ public class MapFile extends MapDataStore {
     private byte zoomLevelMin = 0;
     private byte zoomLevelMax = Byte.MAX_VALUE;
 
+    // container for loaded tags
+    private final ArrayList<Tag> tags = new ArrayList<>();
```
devemux86, line 192: Are you sure we won't have any race condition with the clear / fill of an external list?

devemux86:
> I use simple String.hashcode to get the "unique" int value. This works without any known problem!

We already use this in the `hashCode` methods of various classes.
e.g. previous implementation of `Tag.hashCode()`

Please see some comments in the code, otherwise everything looks good without issues in the tests.

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
tqu-dev: I thought this too. Especially when using multiple rendering threads.

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/StandardRenderer.java
menion: Good point, I'll improve it.

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
menion: Ah. does it mean that MF may read from `MapFile` from multiple threads at once? If you confirm it for me, I'll revert the improvement no 2. , because it lost its benefit in the reduced number of created lists.

menion:
I've just committed a reverted version of the List vs Array improvement (improvement no 2.). So final result is reduced time to 82% based on my tests.

All should be tested and ready to merge as is. Thanks for confirmation on your side!

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
```diff
@@ -27,7 +27,11 @@
 import org.mapsforge.core.util.LatLongUtils;
 import org.mapsforge.core.util.MercatorProjection;
 import org.mapsforge.core.util.Parameters;
-import org.mapsforge.map.datastore.*;
+import org.mapsforge.map.datastore.MapDataStore;
+import org.mapsforge.map.datastore.MapReadResult;
+import org.mapsforge.map.datastore.PoiWayBundle;
+import org.mapsforge.map.datastore.PointOfInterest;
+import org.mapsforge.map.datastore.Way;
```
devemux86, line 34: Can you revert this optimization of imports?
(as the code is unchanged to have clean history)

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/StandardRenderer.java
```diff
@@ -17,7 +17,11 @@
  */
 package org.mapsforge.map.layer.renderer;
 
-import org.mapsforge.core.graphics.*;
+import org.mapsforge.core.graphics.Bitmap;
+import org.mapsforge.core.graphics.Display;
+import org.mapsforge.core.graphics.GraphicFactory;
+import org.mapsforge.core.graphics.Paint;
+import org.mapsforge.core.graphics.Position;
```
devemux86, line 24: @menion
Can you revert this optimization of imports?
(as the code is unchanged to have clean history)

devemux86:
@menion thanks, it looks good.

Minor to have clean history, please see the comments in 2 classes with unchanged code.

menion:
I thank you!

PR #1371, opened 2022-12-05
Support for curved lines

menion:
Lines, mainly contours etc. suffer from straight lines, as seen on the screenshot.
image

This PR brings support for the new attribute "curve" for lines. With the value "cubic" the result is following
image

If there are missing some updates in docs, please let me know, thanks.

### DIFF
--- a/resources/renderTheme.xsd
+++ b/resources/renderTheme.xsd
@@ -42,6 +42,13 @@
         </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="curve">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="cubic" />
+            <xs:enumeration value="no" />
+        </xs:restriction>
+    </xs:simpleType>
+
     <xs:simpleType name="display">
         <xs:restriction base="xs:string">
             <xs:enumeration value="always" />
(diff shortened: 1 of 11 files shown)


resources/renderTheme.xsd
```diff
@@ -188,6 +195,7 @@
         <xs:attribute name="stroke-dasharray" type="tns:strokeDasharray" use="optional" />
         <xs:attribute name="stroke-linecap" default="round" type="tns:cap" use="optional" />
         <xs:attribute name="stroke-linejoin" default="round" type="tns:linejoin" use="optional" />
+        <xs:attribute name="curve" type="xs:string" use="optional" />
```
devemux86, line 198: We probably need to add `default="no"`.

devemux86:
Thanks, nice work!

How is the performance when is enabled?

menion:
I'm testing it now by simply changing the default from "no" to "cubic". This, in my favorite zoom level 12 near Prague, means a huge amount of ways. All will curve=cubic causes a reduction in the draw by 6%. Nothing serious I believe. It may be interesting to test it in the Alps with a lot of contour lines visible. There may be a noticeable impact I believe.

devemux86:
@menion thanks!

PR #1372, opened 2022-12-08
Option to transform text before rendering

menion:
New tag that allows to transform text before rendering.

**Tag**: "text-transform"

**Options**:
- "none" > default, no change
- "uppercase" > all letters uppercase
- "lowercase" > all letters lowercase
- "capitalize" > first letter capitalized

**Usable for**:
- area caption
- POI caption
- way text

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/RenderInstruction.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/RenderInstruction.java
@@ -20,6 +20,7 @@ package org.mapsforge.map.rendertheme.renderinstruction;
 import org.mapsforge.core.graphics.Bitmap;
 import org.mapsforge.core.graphics.GraphicFactory;
 import org.mapsforge.core.graphics.Position;
+import org.mapsforge.core.graphics.TextTransform;
 import org.mapsforge.core.model.Rectangle;
 import org.mapsforge.core.util.Parameters;
 import org.mapsforge.map.datastore.PointOfInterest;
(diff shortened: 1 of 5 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/RenderInstruction.java
```diff
@@ -173,4 +175,22 @@ protected Scale scaleFromValue(String value) {
...
+    protected String transformText(String text, TextTransform transformation) {
+        if (text.length() == 0) {
+            return text;
+        }
+        if (transformation == TextTransform.UPPERCASE) {
+            return text.toUpperCase();
+        } else if (transformation == TextTransform.LOWERCASE) {
+            return text.toLowerCase();
+        } else if (transformation == TextTransform.CAPITALIZE) {
+            return text.substring(0, 1).toUpperCase() + text.substring(1);
```
devemux86, line 191: The case conversions are to be done with `Locale.getDefault()`?

menion: Yes, that's right. Do you think there is a better option?

devemux86: Some recommend `Locale.ROOT` in generic case conversions,
however these are local names, so not sure what is best option.

menion: Oki, thanks. Let's give it a try.

devemux86:
@menion thanks!

devemux86:
See https://github.com/mapsforge/vtm/issues/991:
- Implement in VTM themes
- Support Mapsforge themes

PR #1373, opened 2022-12-08
Fix: not working rotate=false for texts along the way

menion:
Currently "rotate=false" attribute is nicely ignored in the `WayDecorator.renderText` method. The default value is `true` and this is also correctly implemented = all texts are rotated to be always text-down. Anyway, in the case of contour lines, this is not wanted behavior and text should be rendered based on the original way direction.

So I made a simple fix that disable automatic rotation for the "false" value, as expected!


devemux86:
@menion thanks!

IgorMagellan:
I liked this improvement.
It is more convenient for travelers when the figures of the heights of the contour lines indicate to him which way the ascent is going. Now it turned out that these labels indicate the direction "downhill".  I think it needs to be changed to "uphill"

PR #1379, opened 2023-01-02
- add: support for 'text-wrap-width' attribute

menion:
Currently, the text is wrapped based on hardcoded parameter 256 (tilesize) * 0.7.

image

The new parameter 'text-wrap-width' for the 'Caption', allows setting maximum width (in pixels) as

Value: 0 ... no wrap 

image

Value > 0 ... wrap 
Example: `text-wrap-width="100"`

image

### DIFF
--- a/resources/renderTheme.xsd
+++ b/resources/renderTheme.xsd
@@ -164,6 +164,7 @@
         <xs:attribute name="position" default="auto" type="tns:position" use="optional" />
         <xs:attribute name="symbol-id" type="xs:string" use="optional" />
         <xs:attribute name="text-transform" default="none" type="tns:textTransform" use="optional" />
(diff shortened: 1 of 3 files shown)


resources/renderTheme.xsd
```diff
@@ -164,6 +164,7 @@
         <xs:attribute name="position" default="auto" type="tns:position" use="optional" />
         <xs:attribute name="symbol-id" type="xs:string" use="optional" />
         <xs:attribute name="text-transform" default="none" type="tns:textTransform" use="optional" />
+        <xs:attribute name="text-wrap-width" default="none" type="xs:integer" use="optional" />
```
devemux86, line 167: As it has an integer type, the default value should be a number?

devemux86:
Thanks!

PR #1380, opened 2023-01-04
Throw exception on missing envvar for Android-related builds

moving-bits:
Throw a gradle exception during build, if the required environment variable "ANDROID_HOME" is undefined
(only applies to Android-related builds)


devemux86:
Thanks!

PR #1391, opened 2023-02-14
- add: new upright parameter

menion:
We added the attribute `rotate` to disable the rotation of `PathText` for contour lines. Unfortunately, this parameter doesn't cover all possible situations - especially different orientations of source lines. 

This PR creates a new parameter `upright` with possible values and removes the previous `rotate` parameter.

**auto** - original text rotation when text is always upright (use as a default value)
**right** - rotate the text up when the path segment (where the text is to be placed) is facing east. / right (it's basically the latest implementation of the rotate attribute in mode rotate=false )
**left** - rotate text up when the path segment the text goes to the west/left

Inspiration based on Mapnik upright attribute. https://github.com/mapnik/mapnik/wiki/TextSymbolizer

Left
image

Right
image

Auto (default)
image

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Upright.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Upright.java
@@ -1,0 +1,29 @@
+package org.mapsforge.core.graphics;
+
+public enum Upright {
+
+    /**
+     * Automatic map rotation.
+     */
+    AUTO,
+
+    /**
+     * Rotate the text up when the path segment (where the text is to be placed) is facing east. / right.
+     */
+    RIGHT,
+
+    /**
+     * Rotate text up when path segment the text goes to the west/left.
+     */
+    LEFT;
+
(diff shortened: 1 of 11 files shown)


devemux86:
@menion thanks!

Can you check the failed build, seems from a test class?

menion:
@devemux86 ah thanks! Fixed

devemux86:
> removes the previous rotate parameter

What about external themes like OpenAndroMaps Elevate?
They have the "rotate" parameter, will they parse and work?

menion:
My bad @devemux86 sorry. I lived in a world where the "rotate" parameter was not public so its full removal was not a problem ... which was of course incorrect.

Fixed in the latest commit. `rotate` is now fully supported and converted during loading into a new extended `Upright=RIGHT` parameter.

devemux86:
> `rotate` is now fully supported and converted during loading into a new extended `Upright=RIGHT` parameter.

@menion thanks!
It is nice to maintain compatibility when possible. :+1:

PR #1392, opened 2023-02-14
Improved 'text-wrap-width' parameter

menion:
Current implementation has two minor issues:

1. 'text-wrap-width' parameter ignores device DPI and is used directly.

2. breaking to lines is not too clever and Android 29+ provide improved version of the `StaticLayout` class with [optional break strategy parameter](https://developer.android.com/reference/android/graphics/text/LineBreaker#BREAK_STRATEGY_HIGH_QUALITY). This allows to apply a more clever logic when separating text into multiple lines.

Example
full name: "Králove-Krnišov v Pardubicích". Max width set to 100*DPI

SIMPLE (default, current)
image

HIGH_QUALITY
image

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Caption.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Caption.java
@@ -185,7 +185,7 @@ public class Caption extends RenderInstruction {
             } else if (TEXT_TRANSFORM.equals(name)) {
                 this.textTransform = TextTransform.fromString(value);
             } else if (TEXT_WRAP_WIDTH .equals(name)) {
-                int maxWidth = XmlUtils.parseNonNegativeInteger(name, value);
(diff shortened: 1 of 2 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Caption.java
```diff
@@ -185,7 +185,7 @@ private void extractValues(GraphicFactory graphicFactory, DisplayModel displayMo
             } else if (TEXT_TRANSFORM.equals(name)) {
                 this.textTransform = TextTransform.fromString(value);
             } else if (TEXT_WRAP_WIDTH .equals(name)) {
-                int maxWidth = XmlUtils.parseNonNegativeInteger(name, value);
+                int maxWidth = (int) (XmlUtils.parseNonNegativeInteger(name, value) * displayModel.getScaleFactor());
```
devemux86, line 188: Very nice!

devemux86:
@menion thanks!

PR #1393, opened 2023-02-24
Improve orientation system for symbols & texts

menion:
Firstly small fix for the "right" symbol orientation and later ...

# Major compare to the current master branch.

1. removed new (not yet published) "upright" parameter.

2. Introduced a new "symbol-orientation" parameter for symbol rotation along the line.

3. Introduced a new "text-orientation" parameter for text rotation along the line.

4. Heavily simplified overcomplicated `WayDecorator.renderSymbol` system, that now correctly handles `repeatStart` (was applied on every segment) and `repeatGap` (with value 0 caused overflow and crash!).

Inspiration comes from existing Mapnik implementation (related PR [here](https://github.com/mapnik/mapnik/pull/2621)).

## "symbol-orientation" parameter has a new bunch of options.

The original symbol I play with (below). The line is a cliff heading from bottom-left to the top-right.
image

auto
symbol_auto

auto_down
symbol_auto_down

left
symbol_left

right
symbol_right

up
symbol_up

down
symbol_down

## "text-orientation" has a limited set compared to symbols

auto
text_auto

auto_down
text_auto_down

left
text_left

right
text_right

# Extra info

- automatic rotation of symbols was incorrect before (as well as "RIGHT" implementation. As visible (here)[https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/WayDecorator.java#L71], the symbol rotates along the line only if "doInvert" is set to "true". Otherwise, the symbol does not rotate at all (equal to the "UP" parameter).

- because of the problem before, backward compatibility for symbols can't be preserved. New defaults are introduced. For symbols it is "RIGHT", so they always rotate along the line. For "texts" auto, so they are always readable.

Small sorry for another iteration of the already improved system. As we work on new themes, some new requirements appear. Thanks for understanding. Feedback is welcome of course.


devemux86:
@menion thanks for the detailed description!

Please take your time to complete the changes and I will see it.
Also please add the Mapsforge license header in the new classes.

> backward compatibility for symbols can't be preserved

Can you explain what you mean here?
Is it about an existing theme rule or internal rendering behavior?

menion:
We've tested it during next last few days and all seems to work correctly.

Headers added.

**Backward compatibility**

In my previous PR was a serious issue. Check the old code for drawing symbols [here](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/WayDecorator.java#L57). In case, `doInvert` is `true`, then `theta` value was computed. Otherwise, it was set to 0 or even less logical previous value! 

But the older implementation was anyway working correctly. 
The default value for `LineSymbol:rotate` parameter was "true", so now it is set to `SymbolOrientation.LEFT`, and in the case of `rotate=false`, now it is set to `SymbolOrientation.UP`.

For texts along the way, all is set correctly.

Thanks for the push. All should be set correctly now.

I'll also ask Tobias (OpenAndroMaps), to very with the next app Beta version that all is working as expected.

devemux86:
@menion thanks for the nice work!

PR #1395, opened 2023-03-14
Fix of "position" parameter on Android API29+

menion:
On modern Android where is used new StaticLayout.Builder to wrap a texts, is not used correctly prepared `Layout.Alignment` parameter. 

This solves this small fix.

### Confirmation

ABOVE_LEFT
image

ABOVE
image

ABOVE_RIGHT
image

LEFT
image

CENTER
image

RIGHT
image

BELOW_LEFT
image

BELOW
image

BELOW_RIGHT
image


devemux86:
@menion thanks!

PR #1447, opened 2023-11-09
Remove FrameBufferOld and FrameBufferHA2

bailuk:
FrameBufferHA3 is now the default for 3 years. It should be tested enough.
Lets remove old frame buffer implementations for better maintainability.


devemux86:
@bailuk thanks!

PR #1490, opened 2024-06-29
Layer API: change exceptions to warnings

moving-bits:


mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
```diff
@@ -23,11 +23,15 @@
...
 import org.mapsforge.map.model.DisplayModel;
 
+import java.util.logging.Logger;
+
 public abstract class Layer {
     protected DisplayModel displayModel;
     private Redrawer assignedRedrawer;
     private boolean visible = true;
 
+    private static final Logger LOGGER = Logger.getLogger(Layers.class.getName());
```
devemux86, line 33: We should use `Layer.class.getName` here.

mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
```diff
@@ -169,7 +173,7 @@ protected void onRemove() {
 
     final synchronized void assign(Redrawer redrawer) {
         if (this.assignedRedrawer != null) {
-            throw new IllegalStateException("layer already assigned");
+            LOGGER.warning("layer already assigned");
```
devemux86, line 176: Probably it is better to `return;` after the warning?

mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
```diff
@@ -178,7 +182,7 @@ final synchronized void assign(Redrawer redrawer) {
 
     final synchronized void unassign() {
         if (this.assignedRedrawer == null) {
-            throw new IllegalStateException("layer is not assigned");
+            LOGGER.warning("layer is not assigned");
```
devemux86, line 185: Same here, probably it is better to `return;` after the warning?

devemux86:
@moving-bits thanks! Please see my comments.

By the way your fork is not rebased on the latest `master` branch?
The checks have passed in this pull request, but it could cause conflicts in other ones. :slightly_smiling_face:

mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
moving-bits: c&p error, fixed

mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
moving-bits: wasn't sure about what is more important - `onAdd()` getting called or directly exiting on invalid state - but added `return` now

mapsforge-map/src/main/java/org/mapsforge/map/layer/Layer.java
moving-bits: similar to above

devemux86:
@moving-bits thanks!

PR #1520, opened 2024-08-20
Hill shading improvements + new algorithm

Sublimis:
- [x] Enable sub-classing of hill shading algorithms, so users of the library can create theirs
- [x] Get rid of the memory mapping of DEM files, use a simple and more efficient `BufferedInputStream` instead (`InputStream `in general, instead of the `ByteBuffer`)
- [x] Create a new hill shading algorithm with better appearance and more customization

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2022 usrusr
+ * Copyright 2024 Sublimis
  *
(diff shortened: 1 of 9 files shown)


devemux86:
@Sublimis thanks!

I am trying to merge the pull request locally to test it, but unfortunately it does not work:

- [Pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests) are best to use topic branches (rebased on latest master)
- There are 2 commits with a description that does not explain their purpose
- There should only be only 1 commit so we can check quickly the changes
- The new class has problems with rebase breaking git, does it have strange line endings?

Sublimis:
Yes, you're right, the new class is using CRLF 🤦. Will try again 👍

PR #1521, opened 2024-08-20
Hill shading improvements + new algorithm #2

Sublimis:
- Enable sub-classing of hill shading algorithms, so users of the library can create theirs

- Get rid of the memory mapping of DEM files, use a simple and more efficient `BufferedInputStream` instead (`InputStream` in general, instead of the `ByteBuffer`)

- Create a new hill shading algorithm with better appearance and more customization

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFileAndroidContent.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2022 usrusr
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -15,6 +16,7 @@
 package org.mapsforge.map.android.hills;
 
 import android.content.ContentResolver;
+
 import org.mapsforge.core.util.IOUtils;
 import org.mapsforge.map.layer.hills.DemFile;
 import org.mapsforge.map.layer.hills.DemFileFS;
@@ -22,11 +24,10 @@ import org.mapsforge.map.layer.hills.DemFileFS;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.ByteBuffer;
 
 public class DemFileAndroidContent implements DemFile {
-    private final DemFolderAndroidContent.Entry entry;
-    private final ContentResolver contentResolver;
+    protected final DemFolderAndroidContent.Entry entry;
+    protected final ContentResolver contentResolver;
 
     public DemFileAndroidContent(DemFolderAndroidContent.Entry entry, ContentResolver contentResolver) {
         this.entry = entry;
@@ -49,7 +50,7 @@ public class DemFileAndroidContent implements DemFile {
     }
 
     @Override
-    public ByteBuffer asByteBuffer() throws IOException {
+    public InputStream asStream() throws IOException {
         InputStream stream = null;
         try {
             String nameLowerCase = entry.name.toLowerCase();
@@ -57,7 +58,7 @@ public class DemFileAndroidContent implements DemFile {
             if (nameLowerCase.endsWith(".zip")) {
                 return DemFileFS.tryZippedSingleHgt(entry.name, stream);
             } else {
-                return DemFileFS.streamAsByteBuffer(entry.name, stream, (int) entry.size);
+                return DemFileFS.streamReadPart(entry.name, stream, (int) entry.size);
             }
         } finally {
             IOUtils.closeQuietly(stream);
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFolderAndroidContent.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/hills/DemFolderAndroidContent.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2022 usrusr
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -19,6 +20,7 @@ import android.content.Context;
 import android.database.Cursor;
 import android.net.Uri;
 import android.provider.DocumentsContract;
+
 import org.mapsforge.core.util.IOUtils;
 import org.mapsforge.map.layer.hills.DemFile;
 import org.mapsforge.map.layer.hills.DemFolder;
@@ -134,7 +136,7 @@ public class DemFolderAndroidContent implements DemFolder {
         }
     }
 
-    static class Entry {
+    public static class Entry {
         final Uri uri;
         final String name;
         final boolean isDir;
@@ -178,11 +180,11 @@ public class DemFolderAndroidContent implements DemFolder {
         };
 
         Cursor c = null;
+        List<Entry> result = new ArrayList<>();
         try {
             c = contentResolver.query(childrenUri, columns, null, null, null);
 
-            List<Entry> result = new ArrayList<>();
-            while (c.moveToNext()) {
+            while (c != null && c.moveToNext()) {
                 String fileDocId = c.getString(0);
                 String name = c.getString(1);
                 String mimeType = c.getString(2);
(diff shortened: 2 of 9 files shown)


devemux86:
@Sublimis thanks!

No shade seems to work only on Android. Desktop still always shows shades (see desktop image).

Is there an easy way for the other algorithms to not show a shade when there is no slope?

Desktop

Sublimis:
This doesn't seem like an effect of the new algorithm (ClearAsymmetryShadingAlgorithm), but something else. Not sure, though, will check and get back to you. Can you compare slightly more zoomed in screenshots with those of other algorithms? If the effect is visible only over the sea, this should be indicative.
Can it be caching effect, i.e. outside areas (those darker) are cached tiles made with other algorithm which shaded them?

Unfortunately there's no easy way for other algorithms (Simple and Diffuse), as they split the shading range in the middle, and use lower half (less shading) for one direction, and upper half (more shading) for other direction. Flat areas being in the middle always get 50% shading. When this is fixed, i.e. we put flat areas at zero shading, and distribute the shading range asymmetrically over two slope directions, we basically arrive at the new algorithm presented (ClearAsymmetryShadingAlgorithm).

devemux86:
@Sublimis thanks for the extra info!

I make the tests without permanent cache, so it is always new rendered tiles.

The sea has different colors with all algorithms on desktop.
The other 2 algorithms have a dark shade, the new algorithm has a light shade.
Maybe there is something on desktop Java color rendering.

devemux86:
> No shade seems to work only on Android. Desktop still shows shades (see desktop image).

This seems to be related to `AwtCanvas.selectHillShadingComposite`,
using `AlphaComposite` (current implementation) or `AwtLuminanceShadingComposite` (no sea shades).

Sublimis:
The solution seems to be very simple. Take a look at line 49 of the `AwtCanvas.java`:

```java
private static final java.awt.Color NEUTRAL_HILLS = new java.awt.Color(127, 127, 127);
```

This is 50% gray shade.

Instead, it should be:

```java
private static final java.awt.Color NEUTRAL_HILLS = new java.awt.Color(0, 0, 0, 0);
```

(last extra zero is alpha component).

Even better than "drawing" a 100% transparent bitmap would probably be to draw nothing at all in this case, lines 483 to 496:
```java
        if (bitmap == null) {
            // apply flat shading
            if (tileRect != null) {
                this.graphics2D.setClip(
                        (int) Math.round(tileRect.left), (int) Math.round(tileRect.top),
                        (int) Math.round(tileRect.right - (int) tileRect.left), (int) Math.round(tileRect.bottom) - (int) Math.round(tileRect.top) // subtract in after rounding to get same error as on neighbor tile
                );
            }
            this.graphics2D.setColor(NEUTRAL_HILLS);
            this.graphics2D.fillRect(0, 0, getWidth(), getHeight());
            this.graphics2D.setComposite(oldComposite);
            this.graphics2D.setClip(null);
            return;
        }
```

devemux86:
@Sublimis thanks for the ideas.

Have you tried them? Unfortunately none of them work.
Only `AwtLuminanceShadingComposite` seems to work.

devemux86:
@Sublimis thanks for your nice work! :+1:

Sublimis:
> Have you tried them? Unfortunately none of them work. Only `AwtLuminanceShadingComposite` seems to work.

Sorry, no, just looking at the code, should have mentioned that.

Sublimis:
@devemux86 
50% gray for "neutral hills" seemed a logical, if not the only possible culprit, and since this constant hasn't changed, are you sure that this problem is fixed for all use cases? What if someone puts `Parameters.LUMINANCE_COMPOSITE = false`, what motivation and consequences would be in doing that?

Sublimis:
@devemux86
If a v0.22.0 release is planned soon, maybe you can give us a few more days for further hill shading improvements? The plan is to remove the gray shade of flat areas from the Simple and Diffuse algorithms without major impact on the algorithms themselves. Maybe then there will be no need for `Parameters.LUMINANCE_COMPOSITE`.

If we do this, is it better to mark them as `@Deprecated` and create new classes with similar names, or to just modify the existing algorithms as such?

devemux86:
@Sublimis 

There is still no plan for the next version, so we have plenty of time for improvements.

> The plan is to remove the gray shade of flat areas from the Simple and Diffuse algorithms without major impact on the algorithms themselves.

That would be great! :+1: 

> Maybe then there will be no need for Parameters.LUMINANCE_COMPOSITE?

If the old algorithms can be improved to work without gray shade on flat areas
and we manage to fix the desktop `AlphaComposite` then we can revert this flag.

>  is it better to mark them as @Deprecated and create new classes with similar names, or to just modify the existing algorithms as such?

Removing the gray shade in flat areas is something we want for all algorithms?
So we can directly modify the existing algorithms.

devemux86:
@Sublimis 

Is there any news on the work to remove the gray shade of flat areas from the Simple and Diffuse algorithms?

devemux86:
@Sublimis 

Is there any news on the work to remove the gray shade of flat areas from the Simple and Diffuse algorithms?

devemux86:
@Sublimis 

Is there any news on the work to remove the gray shade of flat areas from the Simple and Diffuse algorithms?

Sublimis:
Yes -- sorry for the wait. This was the easy part, but we also created two new algorithms, both of which can look similar to the diffuse, but with more possible adjustments and optimizations, and clearer math. One of them is the "ultimate" high resolution algorithm (explanation comes with the code), and for the last week we've been trying to parallelize it to improve performance. This was a bit of a nightmare, but we succeeded yesterday. Multi-threading as an option will therefore be available in our other algorithms as well.

Since more work is needed on the new algorithms, we can create a pull request just for the fixes to the existing algorithms to reduce the delay, and then another later for the new algorithms, is that okay?

Here is an example of the original versus the corrected diffuse algorithm as it currently stands:

20240905_diffuse_original20240905_diffuse_corrected

Sublimis:
And the high resolution one as it currently stands:

20240905_high_resolution

devemux86:
@Sublimis Impressive results! :+1: 

Which algorithms will have a multi-threading option?
The new ones or all algorithms?

> Since more work is needed on the new algorithms, we can create a pull request just for the fixes to the existing algorithms to reduce the delay, and then another later for the new algorithms, is that okay?

Let's do this, so we can test the changes in the old algorithms.

> high resolution

Do you use 1" for this? Aren't they slow to process?

Sublimis:
We ended up fighting on two fronts: While working on our shading algorithms, we found some inefficiencies in `HgtCache`, the class that prepares hill shading data for rendering, due to a lack of proper synchronization. As if someone forgot that e.g. `HgtCache.getHillshadingBitmap()` can be called simultaneously by multiple threads. This could (and did, as debugging clearly showed) lead to race conditions i.e. extra redundant work being done by more than one thread (for example indexing HGT files, but also bitmap rendering), a work that ends up being discarded for all but the thread that finishes first.

So we took care of this, and while there, we also implemented reading DEM data from ZIP archives containing multiple HGT files, so files downloaded from e.g. viewfinderpanoramas.org can be used as is, without repackaging.

Should we make a separate pull request for these improvements (separate from hill shading algorithms)?

devemux86:
@Sublimis 

> Should we make a separate pull request for these improvements (separate from hill shading algorithms)?

Let's do it this way. :+1:

devemux86:
@Sublimis 

One thing I have seen with the recent improvements, is that the hill shading is not as "strong" anymore.

For example, in the past the "simple" algorithm had more "strong" hill shading, it looked better to the eyes.

devemux86:
@Sublimis 

One thing I have seen with the recent improvements, is that the hill shading is not as "strong" anymore.

For example, in the past the "simple" algorithm had more "strong" hill shading, it looked better to the eyes.
(using the same 3" data)

Sublimis:
Unfortunately, there is no single "perfect" method of shading hills, it's all a matter of wishes and compromises. Let us explain.

Previous versions of the algorithms had maximum contrast between NW and SE facing slopes: if you imagine a very steep hill with a triangle-like cross-section, spanning from SW to NE, then the NW-facing slope would have a shading value of 0, and the SE-facing slope would have 255 (in the range 0-255). At the top of such a hill there would be a "cut", an apparent line where two such opposite slopes meet, one with 0 and the other with 255 shading, so we can call it a "maximum contrast cut". Walking over the top of such a hill, the shading quickly changes from 0 to 127 (horizontal) to 255. However, for this to happen, the horizontal surface must be between the two extremes (unwanted horizontal shading), and the shading of NW slopes decreases as the slope is steeper.

The current versions do not have such a maximum cut at the top of the hill, but a milder one. Horizontal surfaces are at 0 shading. NW slopes get shading proportional to the steepness, up to some maximum that can be less than 255 (and by default it is; 128 in the case of simple shading, which would be called 50% asymmetry in light of upcoming algorithms). The shading of SE slopes is similar, but always goes to 255. Walking over the top of such a hill, the shading quickly changes from 128 to 255 (going down to 0 and back up in an instant), which is less contrast.

This can be alleviated somewhat (in upcoming algorithms, but perhaps also in the current ClearAsymmetry implementation) by adjusting the algorithm parameters, for example the maximum slope and the asymmetry factor.

PR #1524, opened 2024-08-21
Render themes: exclusive / except negation

devemux86:
A minus sign "`-`" excludes the other values after "`|`". It never works alone.

For example:
```xml
<!-- match tunnel-tag that are not 'no' or 'false' -->
<rule k="tunnel" v="-|no|false">
```


demiantres:
The same could also be added in VTM?

devemux86:
@demiantres 

> The same could also be added in VTM?

It already exists in VTM and works the same way.

PR #1534, opened 2024-09-06
Hill shading algorithms improvements

Sublimis:
* Remove gray shading from flat areas in Simple and Diffuse algorithms
* Modify the test for the Diffuse algorithm so it can pass
* Misc. smaller improvements


devemux86:
@Sublimis thanks!

PR #1537, opened 2024-09-15
Hill shading improvements

Sublimis:
* Fix inefficiencies due to lack of proper synchronization in `HgtCache` that lead to race conditions and extra redundant work being done by more than one thread
* Fix inefficiencies due to lack of proper synchronization in `MemoryCachingHgtReaderTileSource` that lead to race conditions and extra redundant work being done and/or redundant objects created by more than one thread
* Ability to use DEM data directly from ZIP archives containing many HGT files; also implemented for Android SAF (storage access framework)
* Misc. smaller improvements


devemux86:
@Sublimis thanks!

PR #1543, opened 2024-09-19
Hill shading new algorithms and improvements

Sublimis:
- New multi-threaded hill shading algorithms
- General hill shading improvements

- Delete mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/ClearAsymmetryShadingAlgorithm.java
- Update HillshadingMapViewer.java
- Update Samples.java


devemux86:
@Sublimis thanks for the detailed work!

The results are indeed impressive. :+1:

devemux86:
@Sublimis 

Is there more work or can I prepare a release for Mapsforge?

Sublimis:
@devemux86 

For now, there are no immediate plans for more work. A little later, maybe in a week or two, we'll try to see if anything else regarding hill shading can be optimized for performance, but for now everything looks fine.

Perhaps it should be noted that we have not published the high-resolution algorithm yet, we'll leave that also for another time. We concluded that it doesn't make much of an improvement over the standard algorithm as it currently stands, first we'll have to improve the equations, if it will be possible.

Thanks! 🙂

devemux86:
@Sublimis 

Is the high-resolution algorithm intended for all data resolutions or only for 1" data?

1" data may be more detailed, but it takes much longer to process and render.
Also they are not available for all countries.

Sublimis:
@devemux86 

It is intended for all DEM resolutions, just like other algorithms (or better to sad "should be intended", if we manage to get it right and publish it).

Yes, the 1" DEM data is 9x larger (25 MB per file), thus one would expect 9x times more processing time, however if the bottleneck of the system is (single-threaded) computing (and not storage or something else) then one can set the number of computing threads (using `ClasyParams`) to e.g. two or three (2 or 3) and improve the performance to a factor less than 9.

devemux86:
> can set the number of computing threads (using ClasyParams) to e.g. two or three (2 or 3) and improve the performance to a factor less than 9.

I tried this, but the result of processing 1" is still slow.
So users will definitely complain about the slow map performance.

devemux86:
@Sublimis 

> can set the number of computing threads (using ClasyParams) to e.g. two or three (2 or 3) and improve the performance to a factor less than 9.

I tried this, but the result of processing 1" is still slow.
So users will definitely complain about the slow map performance.

Sublimis:
@devemux86 

You're testing on desktop? We've tested only on Android, but on desktop it should be even faster, at least in theory.
Older algorithms (Diffuse, Simple), are also slow (and even slower), right?

If the bottleneck is the storage, then there's nothing you can do to improve performance unfortunately.
You can override the `notifyReadingPaced(int)` callback and see if it ever gets called (with 1 computing thread, the default). If not, then the storage is a definite bottleneck on your test system.

devemux86:
@Sublimis 

I test both platforms, but 1" is slow even on powerful desktop with ssd compared to 3".
(and it is slower on mobile)

Sublimis:
@devemux86 

But the new algorithm is not slower in this regard compared to older algorithms, when multi-threading is used?

devemux86:
@devemux86 

I just mean that I tried the new multi-threading with 1" and have not seen significant improvement.
(changing `AThreadedHillShading.ComputingThreadsCountDefault` to e.g. 8)

devemux86:
@Sublimis 

I just mean that I tried the new multi-threading with 1" data and have not seen significant improvement.
(changing `AThreadedHillShading.ComputingThreadsCountDefault` to e.g. `8`)

devemux86:
@Sublimis 

The new improved hill shade rendering is more clear, very nice work! :+1:

Sublimis:
@devemux86 

Thanks! We really optimized the calculations heavily, so you probably do not get `notifyReadingPaced` called because calculations are much faster than reading data on most systems, even with just one extra computing thread. Maybe in the future it would be good to see if DEM data reading can be made multi-threaded as well (currently only one thread is reading the data). Not sure if this would help, but who knows.

devemux86:
@Sublimis 

I also like the `HillShadingUtils`.
We use something similar in VTM, see the [FastMath](https://github.com/mapsforge/vtm/blob/master/vtm/src/org/oscim/utils/FastMath.java) and [MathUtils](https://github.com/mapsforge/vtm/blob/master/vtm/src/org/oscim/utils/math/MathUtils.java).

---

How many computing threads do you recommend?
Depending on device cpu cores? Or keep the default 1?

Sublimis:
@devemux86 

The default value of 1 is sufficient for most systems, since the calculations are few and fast. Also don't forget that the `AbsShadingAlgorithmDefaults .convert()` method is already being called from multiple threads at the same time. We'll try to compile the test results next week, but as a theoretical example, let's assume for simplicity that the reading task takes 80% of the time, while the calculation task takes 20%. A multi-threaded implementation will theoretically perform computing ("20% task") at the same time as reading ("80% task"), so the total runtime will be 80% of the time of a single-threaded implementation. More computing threads are only really needed if the computation time increases to more than 50% of the total runtime in a single-threaded implementation. This is of course difficult to assess, so if someone is interested in whether more threads should be inserted into the algorithm, they should see if `notifyReadingPaced` is called and how many times.

devemux86:
@Sublimis 

Have you noticed any difference in performance if we use zipped dem data?

Sublimis:
So, we implemented multi-threaded reading of DEM files as well. Speeds up hill shading on systems where a single thread cannot saturate storage. We can prepare the pull request tomorrow.

devemux86:
@Sublimis 

Could we implement an API to query the elevation for specific coordinates from the hill shading data?

Sublimis:
Of course. We're talking about use in navigation, for a given array of lat/lon-s return an array of altitudes, right?

devemux86:
@Sublimis 

For all purposes, something simple: input = (latitude, longitude) and output = elevation.

If it is expensive to retrieve, it could be extended to for list of coordinates.

Sublimis:
@devemux86 

Not a problem, but it may take a week or two to implement. If the release of the new version is planned earlier, it could be postponed to the next one.

Here you can try the fully multi-threaded hill shading, with 2 read threads and 1 compute thread by default (per caller thread): [https://github.com/Sublimis/mapsforge/tree/hillshading-04](https://github.com/Sublimis/mapsforge/tree/hillshading-04). We haven't submitted a pull request yet as we are now experimenting with bicubic interpolation applied to the hill shading, which should theoretically make it look the best it can (smoother). Everything that has been done so far remains valid, this will only be an addition for use cases where looks are more important than performance. We should have the first results by tomorrow, if all goes well the pull request could be ready soon after (with additional modifications to the `AThreadedHillShading.java`).

devemux86:
@Sublimis Thanks for your work, I will check it! :+1: 

About the release of the next version, we should not delay it too much,
as it's important that developers can try the next version soon, it already contains many new features.

Let's complete the hillshading rendering with what else is needed
and we can add the elevation API after the next version.

devemux86:
@Sublimis all seem fine.

- It mentions 2 reading threads, but the `AThreadedHillShading.ReadingThreadsCountDefault` is 1?
Or it means 1 extra reading thread + the main reading thread = 2?

- Please check the formatting of the code.
This branch has several changes with unneeded new lines in `{` and `}`.

devemux86:
@Sublimis 

I will wait for your next pull request with all the improvements
and if there is no more work, I can program a new version of Mapsforge.

Sublimis:
@devemux86 

>     * It mentions 2 reading threads, but the `AThreadedHillShading.ReadingThreadsCountDefault` is 1?
>       Or it means 1 extra reading thread + the main reading thread = 2?

Yes, that means 1 additional thread. We further clarified this in the javadoc.
A pull request with additional changes to `AThreadedHillShading` (and small tweaks to `AClasyHillShading` to support the new functionality) will be ready today.

devemux86:
@Sublimis 

Do you mean it will include all remaining hillshading rendering improvements?

You mentioned above about bicubic interpolation.

Sublimis:
@devemux86 

> @Sublimis
> 
> Do you mean it will include all remaining hillshading rendering improvements?
> 
> You mentioned above about bicubic interpolation.

Yes.
Sorry for the delay. We are working around the clock to complete this as soon as possible.

Sublimis:
Done, PR will be made sometime during the day

PR #1544, opened 2024-09-20
Correct pull of data from MultiMapDataStore

menion:
Recent implementations of the `read` operations in the `MultiMapDataStore` are based on the checking of the upper-left tile only. 

This results in missing data in case of for example low zoom level and multiple maps covered by area defined by "upperLeft" and "lowerRight" tiles.

On the attached screen is an example of "readLabels" results, which should cover the whole screen, but the results are only data from the first DE map.

image

### DIFF
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
@@ -1073,8 +1073,9 @@ public class MapFile extends MapDataStore {
     }
 
     @Override
-    public boolean supportsArea(BoundingBox boundingBox) {
-        return boundingBox.intersects(getMapFileInfo().boundingBox);
+    public boolean supportsArea(BoundingBox boundingBox, byte zoomLevel) {
+        return boundingBox.intersects(getMapFileInfo().boundingBox)
+                && (zoomLevel >= this.zoomLevelMin && zoomLevel <= this.zoomLevelMax);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


devemux86:
@menion thanks!

This seems right, although I cannot reproduce the report with e.g. Mapsforge desktop samples app.
All this time multiple maps seem to render correctly.

How we can trigger this using the Mapsforge samples?

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
```diff
@@ -1072,6 +1072,11 @@ public boolean supportsTile(Tile tile) {
                 && (tile.zoomLevel >= this.zoomLevelMin && tile.zoomLevel <= this.zoomLevelMax);
     }
 
+    @Override
+    public boolean supportsArea(BoundingBox boundingBox) {
+        return boundingBox.intersects(getMapFileInfo().boundingBox);
```
devemux86, line 1077: This should also contain the zoom level check like `supportsTile`.

devemux86:
@menion 

I understand, please see my above 2 comments on the code.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
```diff
@@ -162,7 +162,7 @@ public MapReadResult readLabels(Tile upperLeft, Tile lowerRight) {
         switch (this.dataPolicy) {
             case RETURN_FIRST:
                 for (MapDataStore mdb : mapDatabases) {
-                    if (mdb.supportsTile(upperLeft)) {
+                    if (mdb.supportsArea(upperLeft.getBoundingBox().extendBoundingBox(lowerRight.getBoundingBox()))) {
```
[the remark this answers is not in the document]
menion: Could you be more precise? `Tile.getBoundingBox()` value is cached, so I do not expect and performance negative effect. Anyway if there is an option to do it faster ...

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
menion: My initial ideal was that zoom level should not play the role here, but as `MapFile` has it's defined limits, you are of course correct that it has to be there as well. I'll improve it.

Issue #1497, opened 2024-07-01
Hillshading empty regions and high CPU on 0 lat / lon (Android)


devemux86:
Report for south Great Britain map.

The reason seems to be that on Android it tries to create bitmaps with zero width or height.
On desktop it works without problems.

<img src="https://github.com/mapsforge/mapsforge/assets/3484020/5130aa69-fe1e-47e0-8feb-b7c4866b1989" width="50%"/><img src="https://github.com/mapsforge/mapsforge/assets/3484020/24c05273-dc57-49ad-883e-41ee0debf2cb" width="50%"/>

devemux86:
cc @usrusr

PR #1561, opened 2024-10-31
Major hill shading performance improvements

Sublimis:
* Major performance improvements when using hill shading.
   - Improved hill shaded tiles overlapping.

       Overlapping is now only done when necessary during drawing, not during loading when there is not enough information about the required tiles. As an (worst case scenario) example, imagine we want to display a map tile where 4 hill-shaded tiles meet on large zoom level. The previous algorithm would wastefully load and process 12 shading tiles just to draw this one map tile: 4 required center tiles, but then also 8 "invisible" tiles: 2 tiles north, 2 tiles west, 2 tiles south, and 2 tiles east of the 4 center tiles, 12 tiles total. Now only the 4 center tiles are loaded and overlapped, since that is all that is needed to render the map tile.

   - Performance of the high resolution (i.e. bicubic) hill shading was improved by almost 4x (four times faster).

       Quick benchmarks show the following times to process a single 1" HGT file (that is 25MB or almost 13M points per file):

       - 900 ms (new high resolution algorithm)
       - 550 ms (new standard resolution algorithm)
       
         Comparing this to a previous version:

       - 3200 ms (v0.22.0 high resolution algorithm)
       - 1550 ms (v0.22.0 diffuse algorithm)
       
       Total times are much more drastic. These are times until the map is completely loaded (DEM files were processed one by one):
       
       - 3.9 seconds (new high resolution algorithm, loading 4 tiles)
       - 2.5 seconds (new standard resolution algorithm, loading 4 tiles)
       - 40 seconds (v0.22.0 high resolution algorithm, loading 12 tiles)
       - 19 seconds (v0.22.0 diffuse algorithm, loading 12 tiles)
      
      All algorithms were on default settings (including preprocessing enabled), on an Android device with 4x 2.5GHz cores and 4x 1.9GHz cores.

* Major visual improvements when using hill shading. Removed nearly half a dozen causes of various artifacts on hill shaded tiles. This also resulted in elegant code simplifications and performance improvements.

   - Artifacts due to missing input data and poor preprocessing
   - Artifacts due to hill shading bitmap overlapping misalignments
   - Artifacts due to hill shading bitmap filtering
   - Artifacts due to poor thread synchronization while drawing hill shading bitmaps

* Input data preprocessing is now a separate step, and an option (enabled by default). Preprocessing removes invalid values from the input data by filling some more clever value, like an average of neighboring points. The preprocessing algorithm is now smarter than before, previously it just filled a fixed neighboring value without checking that it also might be invalid. It is now also being done during processing phase, not reading phase, i.e. on a computing thread pool (thus multi-threaded).

* Major optimization and performance improvements to AwtCanvas.shadeBitmap().
   
   - Preventing needless up-scaling of the entire hill shading bitmap, that previously wasted large amounts of memory and CPU time when only a small part of the bitmap is really needed. This was especially prominent on larger zoom levels, and could potentially cause an OOM exception.

Examples of artifacts that were successfully handled:

2024-10_artifact1r 2024-10_artifact2r 2024-10_artifact3r


devemux86:
@Sublimis Thanks, quick question as there changes in the 2 Canvas classes:

There was a report about broken hillshading and high CPU at 0 latitude and 0 longitude:
- https://github.com/mapsforge/mapsforge/issues/1497

This was fixed with changes in the 2 Canvas classes:
- https://github.com/mapsforge/mapsforge/pull/1498

Have you tested the new code at 0 latitude and 0 longitude?

Sublimis:
> @Sublimis Thanks, quick question as there are changes in the 2 Canvas classes:
> 
> There was a report about broken hillshading and high CPU at 0 latitude and 0 longitude:
> 
>     * [Hillshading empty regions and high CPU at 0 lat / lon #1497](https://github.com/mapsforge/mapsforge/issues/1497)
> 
> 
> This was fixed in the 2 Canvas classes:
> 
>     * [Fix hillshading at 0 lat / lon, fix #1497 #1498](https://github.com/mapsforge/mapsforge/pull/1498)
> 
> 
> Have you tested the new code at 0 latitude and 0 longitude?

Yup, no problems whatsoever. That was probably some manifestation of inefficiencies and/or poor synchronization that was taken care of in this PR. Screenshot:

image

devemux86:
@Sublimis 

It produces pixelated images on older Android. Code on `master` works correctly.
Screenshot is with Mapsforge Samples on our min supported Android 5 (API 21):

Sublimis:
Easily fixed with a few lines of code in just `AndroidCanvas`. As it turns out, on Android 7 and older an `ALPHA_8` bitmap (like hill shadings) must be converted to `ARGB_8888` before upscaling, otherwise no filtering will be applied. Why it works in the current master is probably luck, because it uses a different implementation of the `android.graphics.Bitmap.createBitmap()` (inspection of the source code reveals that there are several, not to mention overrides), probably some undocumented magic happens there on old Androids.

Can we do a PR just to patch this up without repeating the previous big PR?
You can inspect the new code here: https://github.com/Sublimis/mapsforge/blob/hillshading-07/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidCanvas.java

devemux86:
@Sublimis You posted an additional commit with changes in AndroidCanvas.
Are more changes needed?

Sublimis:
> Are more changes needed?

No, that's all.

devemux86:
@Sublimis Thanks, I will check it. :+1:

Sublimis:
Oh, sorry, haven't realized that the patch commit immediately appeared in this PR. Can you squash them?

devemux86:
They can be squashed during the merge.

devemux86:
@Sublimis Thanks for these improvements! :+1:

Sublimis:
It was a pleasure 👍

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
```diff
@@ -83,7 +84,7 @@ public DatabaseRenderer(MapDataStore mapDataStore, GraphicFactory graphicFactory
      *
      * @param rendererJob the job that should be executed.
      */
-    public TileBitmap executeJob(RendererJob rendererJob) {
+    public synchronized TileBitmap executeJob(RendererJob rendererJob) {
```
devemux86, line 87: @Sublimis 

Why we need the `synchronized` here?

Sublimis: > @Sublimis
> 
> Why we need the `synchronized` here?

`Canvas` (used in the `CanvasRasterer` throughout the `executeJob()` method) should not be considered thread-safe, at least not on Android. `Canvas` thread-safety assumption was also the reason for one of the artifacts mentioned above, specifically the one in the first image.

devemux86: I wonder if this affects the multi threading vector rendering (not hill shading).
(as we have not seen problems there without the `synchronized` so far)

Sublimis: You mean performance? Performance is probably the only thing that could possibly be affected by synchronizing this method. Haven't noticed any decline in performance though, but will try to make some more objective testing. Can you suggest some especially complex scene that takes notable amount of time to render?

devemux86: Note: I mean the normal map rendering **without** hill shading, i.e. test only the vector maps.

Very populated areas like London at zoom levels where much data is processed /rendered,
with 1 thread we wait the tiles to render, while the (default) multi threading is much faster.

Sublimis: > Very populated areas like London at zoom levels where much data is processed /rendered, with 1 thread we wait the tiles to render, while the (default) multi threading is much faster.

You mean with the current synchronized paradigm there is a noticeable wait that was not present before?
Ok, will let you know after looking into that.

devemux86: Exactly, I mean the current code in `master` branch with and without this `synchronized`,
if it affects the normal map rendering without hill shading when multi threading is enabled.

Sublimis: Done and fixed. Thanks for pointing this out. The cause of artifacts like in the first image was pinpointed to be the concurrent access/modification of hill shading bitmaps through Canvas in very distant code sites, so a proper synchronization of minimum granulation was applied there. Will put the code (containing some other exciting improvements) to our fork tomorrow, so you can check it out before we make a PR.

Sublimis: Sorry for the delay. You can see our latest improvements here: https://github.com/Sublimis/mapsforge/pull/14. The commit is quite intricate, please take your time and ask if anything is unclear. We're quite happy with everything, this nicely rounds off all the work done on hill shading so far, and raises it to another level.

devemux86: @Sublimis Thanks, there are too many changes in the whole library core classes just for hill shading.

Why do we need them?
The description only mentions the hill shading improvements.

Documentation is needed per change and class in the pull request
or multiple steps and pull requests to understand all of them.

The many users of the library are usually interested in the normal vector map rendering.
And if any core changes have positive or negative effect of the normal map performance.

Can you give more details about all other changes?

Also please avoid moving existing methods in other line, it makes review more difficult.
Only new or changed code should exist in commits where there is functionality change.

devemux86:
For now I reverted the `synchronized` in `DatabaseRenderer`
to avoid possible performance issues in normal map rendering.

---

For hillshading improvements, better only change the hillshading classes.

Changes in other classes should be discussed before to clarify their need
and be done separately in small pull requests per case.

This way it is easier to review and understand everything in small steps. :slightly_smiling_face: 

It looks like there are changes not related to hill shading (e.g. layers, labels).
Also `synchronized` and `volatile` in core classes need a good explanation as they _might_ affect performance.

devemux86:
@Sublimis 

For now I reverted the `synchronized` in `DatabaseRenderer`
to avoid possible performance issues in normal map rendering.

---

For hillshading improvements, better only change the hillshading classes.

Changes in other classes should be discussed before to clarify their need
and be done separately in small pull requests per case.

This way it is easier to review and understand everything in small steps. :slightly_smiling_face: 

It looks like there are changes not related to hill shading (e.g. layers, labels).
Also `synchronized` and `volatile` in core classes need a good explanation as they _might_ affect performance.

Sublimis:
We'll try to break down some of the changes in packages other than "layer.hills" as briefly as possible.

This code is called from multiple threads, so it's simply wrong:

https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-core/src/main/java/org/mapsforge/core/model/Tile.java#L222-L234

Even if it worked before, it's a disaster waiting to happen, it can cause bugs that are hard to find. The `getBoundingBox()` was made thread-safe in a similar fashion, as there are no guarantees that it won't be called from multiple threads (it probably is).

This code is also called from multiple threads:

https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/RenderTheme.java#L80-L85
https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/RenderTheme.java#L211-L213

This flag is also accessed from multiple threads:

https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/RendererJob.java#L26C26-L26C26

This mutex is only used for hill shading code, in two places where needed to exclude concurrent access:

https://github.com/Sublimis/mapsforge/blob/05281fdf31b0246eb82a7b2cfcb6c6d544da5e8e/mapsforge-core/src/main/java/org/mapsforge/core/graphics/ABitmap.java#L19-L24

Learning from the situation with the mutex above, this field cannot be guaranteed to be modified from a single thread and should also be considered a disaster waiting to happen (in AWT there's a similar field):

https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidBitmap.java#L65

For this we merely introduced getter and setter for better encapsulation (it appears to be called only from one thread):

https://github.com/mapsforge/mapsforge/blob/452dc3f07aa7dc496342fa9c67ef461a404dbb02/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/RenderContext.java#L47C46-L47C46

`RenderContext.drawingLayers` was simply renamed to a singular form, since it actually holds a single layeR, not multiple layerS so it was very confusing 🥲 

Majority of the code mentioned is in fact somehow related to hill shading, except for these last two things. Tell us if we forgot something?

Also, the fields that have been made volatile are not contended, so this should be relevant: https://stackoverflow.com/questions/4633866/is-volatile-expensive

Thanks!

devemux86:
> We'll try to break down some of the changes in packages other than "layer.hills" as briefly as possible.

Thanks! :+1: 

> This code is called from multiple threads

Why not adding `synchronized` to such methods?
(like we already do in [Layers](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/Layers.java) class)

> Majority of the code mentioned is in fact somehow related to hill shading, except for these last two things. Tell us if we forgot something?

Another example is changes in the constructor of `DatabaseRenderer` about the label layer.

There is no problem to change or improve other classes, but we have to do this separately,
in small steps and pull requests with good documentation about why the change is needed.

So we can easily review them in the future when we work again in these library areas. :slightly_smiling_face:

devemux86:
> We'll try to break down some of the changes in packages other than "layer.hills" as briefly as possible.

Thanks! :+1: 

> This code is called from multiple threads

Why not adding `synchronized` to such methods?
(like we already do in [Layers](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/layer/Layers.java) class)

> Majority of the code mentioned is in fact somehow related to hill shading, except for these last two things. Tell us if we forgot something?

Another example is changes in the constructor of `DatabaseRenderer` about the label layer.

There is no problem to change or improve other classes, but we have to do this separately,
in small steps and pull requests with good documentation about why the change is needed.

So we can easily review them in the future when we work again in these library areas. :slightly_smiling_face:

devemux86:
My suggestion is that if you see issues in the library with multi threading and synchronization,
let's first look at each case individually and possibly improve it through separate pull requests.

When these multi threading issues are resolved we can proceed with the hill shading changes,
if they first need the above multi threading improvements to work.

Sublimis:
> The story went like this. We were hunting down the bug that was causing the artifact on a hill shading, and this required going through the entire rendering pipeline code, basically most of the library. While doing this, we encountered many inconsistencies and/or (multi-threading) issues in many places, most of which were ignored in this PR, and we decided to only make a handful of fixes for the final commit, usually those related to hill shading or memory leaks.

It was important to point this out, since we probably don't want to go through the entire code twice in our lives, it was a "now or never" kind of deal 😄

devemux86:
@Sublimis 

> we encountered many inconsistencies and/or (multi-threading) issues in many places

The library is not perfect and we can all certainly improve it.
But this has to be done step by step in separate pull requests.

One for Tile, one for Bitmap, one for DatabaseRenderer, ... etc.

devemux86:
@Sublimis As you continue to look for improvements in hillshading:

Using the Motorider **dark** theme with hillshading,
the land remains "black" on Android, while on desktop the land changes to "gray".

Sublimis:
> @Sublimis As you continue to look for improvements in hillshading:
> 
> Using the Motorider **dark** theme with hillshading, the land remains "black" on Android, while on desktop the land changes to "gray".

Already taken care of this 💪
Thanks!

image

image

devemux86:
@Sublimis What is this green and brown color?

Sublimis:
> @Sublimis What is this green and brown color?

Hill shading, of course, but this time in a color other than black

PR #1577, opened 2024-11-18
Fix: DEM ZipFile-s not being closed

Sublimis:
This caused a memory leak observed on desktop (wasn't tested on Android). The leak becomes apparent when many HGT files are used for hill shading.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/DemFileZipEntryFS.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/DemFileZipEntryFS.java
@@ -15,6 +15,7 @@
 package org.mapsforge.map.layer.hills;
 
 import java.io.BufferedInputStream;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/DemFileZipEntryFS.java
```diff
@@ -23,31 +24,36 @@
 
 public class DemFileZipEntryFS implements DemFile {
 
-    protected final ZipFile zipFile;
-    protected final ZipEntry zipEntry;
+    protected final File mFile;
+    protected final String mZipEntryName;
+    protected final long mZipEntrySize;
```
devemux86, line 29: @Sublimis Please do not use the `m` field naming, as we do not use it in Mapsforge.

It also overloads the commit history with unrelated changes in the pull request.

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/DemFileZipEntryFS.java
```diff
@@ -23,31 +24,36 @@
...
 public class DemFileZipEntryFS implements DemFile {
 
-    protected final ZipFile zipFile;
-    protected final ZipEntry zipEntry;
+    protected final File zipFile;
+    protected final String zipEntryName;
+    protected final long zipEntrySize;
 
-    public DemFileZipEntryFS(ZipFile zipFile, ZipEntry zipEntry) {
+    public DemFileZipEntryFS(File zipFile, String zipEntry, long zipEntrySize) {
```
devemux86, line 31: `zipEntry` better be named `zipEntryName` to be the same as the class field.

devemux86:
@Sublimis Thanks!

PR #1578, opened 2024-11-19
Fix: Labels overlap when map is rotated

Sublimis:
Related (especially the comment): https://github.com/mapsforge/mapsforge/blob/bbc6425f42f6161fdbf50e5f69cebe1829d1ece7/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPointTextContainer.java#L150-L162

Before/After:

20241119_before20241119_after

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
@@ -1,6 +1,7 @@
 /*
  * Copyright 2010, 2011, 2012, 2013 mapsforge.org
  * Copyright 2014 Ludwig M Brinckmann
+ * Copyright 2024 Sublimis
  *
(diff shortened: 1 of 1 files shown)


Sublimis:
Please note that this fix is not perfect and may require more work, as it doesn't deal with offsets. Every proper solution should probably also recognize the additional offsets added, e.g.:

https://github.com/mapsforge/mapsforge/blob/bbc6425f42f6161fdbf50e5f69cebe1829d1ece7/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPointTextContainer.java#L160-L161

and

https://github.com/mapsforge/mapsforge/blob/bbc6425f42f6161fdbf50e5f69cebe1829d1ece7/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPointTextContainer.java#L200-L201

However, it does appear to be improvement over the current implementation.

devemux86:
@Sublimis Thanks!

PR #1579, opened 2024-11-22
Fix: Labels & co.

Sublimis:
* Fix 🐞: Nondeterministic labels, i.e. certain labels sometimes appear, sometimes not, everything else the same. (This bug has been bothering us for the last cca 10 years!)
* Fix 🐞: Better label overlap avoidance algorithm
* Fix 🐞: Label overlap check not refreshed on map rotation
* Android: Label space now includes a small padding when checking for overlap (avoids placing labels too close together, which can sometimes appear as one word)
* Android: Better handling of multiline labels
* Android: Test for label overlap algorithm

## After:

20241122-after1

20241122-after2

**`PointTextContainer.DEBUG_CLASH_BOUNDS` flag example:**

20241122-after-debug

## Before:

20241122-before1
20241122-before2

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
@@ -2,6 +2,7 @@
  * Copyright 2010, 2011, 2012, 2013 mapsforge.org
  * Copyright 2014 Ludwig M Brinckmann
  * Copyright 2016-2017 devemux86
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -25,7 +26,39 @@ public interface Paint {
 
     int getTextHeight(String text);
 
+    default int getTextHeight(final String text, boolean includePadding) {
+        int retVal = 0;
+
+        retVal = getTextHeight(text);
+
+        if (includePadding && retVal > 0) {
+            retVal += getFontPadding();
+        }
+
+        return retVal;
+    }
+
     int getTextWidth(String text);
+
+    default int getTextWidth(String text, boolean includePadding) {
+        int retVal = 0;
+
+        retVal = getTextWidth(text);
+
+        if (includePadding && retVal > 0) {
+            retVal += getFontPadding();
+        }
+
+        return retVal;
+    }
+
+    default int getTextWidth(final String text, final int widthMax, boolean includePadding) {
+        return getTextWidth(text);
+    }
+
+    default int getFontPadding() {
+        return 0;
+    }
 
     boolean isTransparent();
 
--- a/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
@@ -25,6 +25,8 @@ import org.mapsforge.core.model.Rotation;
 
 public abstract class PointTextContainer extends MapElementContainer {
 
+    protected static final boolean DEBUG_CLASH_BOUNDS = false;
+
     public final double horizontalOffset;
     public final boolean isVisible;
     public final int maxTextWidth;
@@ -33,9 +35,9 @@ public abstract class PointTextContainer extends MapElementContainer {
     public final Position position;
     public final SymbolContainer symbolContainer;
     public final String text;
-    public final int textHeight;
-    public final int textWidth;
     public final double verticalOffset;
+    public volatile double clashRotationDegrees;
+    public volatile Rectangle clashRect;
 
     /**
      * Create a new point container, that holds the x-y coordinates of a point, a text variable, two paint objects, and
@@ -52,13 +54,6 @@ public abstract class PointTextContainer extends MapElementContainer {
         this.paintFront = paintFront;
         this.paintBack = paintBack;
         this.position = position;
-        if (paintBack != null) {
-            this.textWidth = paintBack.getTextWidth(text);
-            this.textHeight = paintBack.getTextHeight(text);
-        } else {
-            this.textWidth = paintFront.getTextWidth(text);
-            this.textHeight = paintFront.getTextHeight(text);
-        }
         this.horizontalOffset = horizontalOffset;
         this.verticalOffset = verticalOffset;
         this.isVisible = !this.paintFront.isTransparent() || (this.paintBack != null && !this.paintBack.isTransparent());
@@ -72,18 +67,21 @@ public abstract class PointTextContainer extends MapElementContainer {
         if (!(other instanceof PointTextContainer)) {
             return false;
         }
+
         PointTextContainer ptc = (PointTextContainer) other;
-        if (!Rotation.noRotation(rotation)) {
-            Rotation mapRotation = new Rotation(rotation.degrees, (float) this.xy.x, (float) this.xy.y);
-            Rectangle rotated = mapRotation.rotate(this.boundary.shift(xy));
-            Rotation otherRotation = new Rotation(rotation.degrees, (float) ptc.xy.x, (float) ptc.xy.y);
-            Rectangle otherRotated = otherRotation.rotate(ptc.boundary.shift(ptc.xy));
-            if (rotated.intersects(otherRotated))
+
+        if (!Rotation.noRotation(rotation) || DEBUG_CLASH_BOUNDS) {
+            Rectangle rect1 = getClashRect(this, rotation);
+            Rectangle rect2 = getClashRect(ptc, rotation);
+
+            if (rect1 != null && rect2 != null && rect1.intersects(rect2)) {
                 return true;
+            }
         }
         if (this.text.equals(ptc.text) && this.xy.distance(ptc.xy) < 200) {
             return true;
         }
+
         return false;
     }
 
@@ -127,5 +125,157 @@ public abstract class PointTextContainer extends MapElementContainer {
         stringBuilder.append(", text=");
         stringBuilder.append(this.text);
         return stringBuilder.toString();
+    }
+
+    /**
+     * @return New rotated x,y position of the container relative to the origin point
+     */
+    protected Point getRotRelPosition(double originX, double originY, Rotation rotation) {
+        return getRotRelPosition(this, originX, originY, rotation);
+    }
+
+    /**
+     * @return New rotated x,y position of the container relative to the origin point
+     */
+    public static Point getRotRelPosition(PointTextContainer pointTextContainer, double originX, double originY, Rotation rotation) {
+        double x = pointTextContainer.xy.x - originX;
+        double y = pointTextContainer.xy.y - originY;
+
+        if (!Rotation.noRotation(rotation)) {
+            Point rotated = rotation.rotate(x, y, true);
+            x = rotated.x;
+            y = rotated.y;
+        }
+
+        // The offsets can only be applied after rotation
+        x += pointTextContainer.horizontalOffset + pointTextContainer.getBoundary().left;
+        y += pointTextContainer.verticalOffset + pointTextContainer.getBoundary().top;
+
+        return new Point(x, y);
+    }
+
+    /**
+     * See org.mapsforge.map.rendertheme.renderinstruction.Caption#Caption,
+     * org.mapsforge.map.android.graphics.AndroidPointTextContainer#AndroidPointTextContainer,
+     * or org.mapsforge.map.awt.graphics.AwtPointTextContainer#AwtPointTextContainer.
+     *
+     * @return Clash rectangle in absolute coordinate space
+     */
+    public static Rectangle getClashRect(PointTextContainer pointTextContainer, Rotation rotation) {
+        if (!pointTextContainer.isVisible) {
+            return null;
+        }
+
+        // All other fields used are final (read only)
+        if (rotation.degrees == pointTextContainer.clashRotationDegrees && pointTextContainer.clashRect != null) {
+            return pointTextContainer.clashRect;
+        }
+
+        final Rotation newRotation = new Rotation(rotation.degrees, 0, 0);
+
+        // We work in the absolute coordinate space of the map (intentionally)
+        double x = pointTextContainer.xy.x;
+        double y = pointTextContainer.xy.y;
+
+        if (!Rotation.noRotation(newRotation)) {
+            Point rotated = newRotation.rotate(x, y, true);
+            x = rotated.x;
+            y = rotated.y;
+        }
+
+        // The offsets can only be applied after rotation
+        x += pointTextContainer.horizontalOffset;
+        y += pointTextContainer.verticalOffset;
+
+        double textOffsetX, textOffsetY;
+
+        // Implementation of the constructors mentioned above allows such use of boundary dimensions
+        textOffsetX = pointTextContainer.getBoundary().getWidth();
+        textOffsetY = pointTextContainer.getBoundary().getHeight();
+
+        final Rectangle output;
+
+        switch (pointTextContainer.position) {
+            case CENTER:
+            default:
+                // Text is positioned centered on x,y
+                textOffsetX /= 2;
+                textOffsetY /= 2;
+                output = new Rectangle(x - textOffsetX, y - textOffsetY, x + textOffsetX, y + textOffsetY);
+                break;
+            case BELOW:
+                // Text is positioned centered-below x,y
+                textOffsetX /= 2;
+                output = new Rectangle(x - textOffsetX, y, x + textOffsetX, y + textOffsetY);
+                break;
+            case ABOVE:
+                // Text is positioned centered-above x,y
+                textOffsetX /= 2;
+                output = new Rectangle(x - textOffsetX, y - textOffsetY, x + textOffsetX, y);
+                break;
+            case BELOW_LEFT:
+                // Text is positioned below-left of x,y
+                output = new Rectangle(x - textOffsetX, y, x, y + textOffsetY);
+                break;
+            case ABOVE_LEFT:
+                // Text is positioned above-left of x,y
+                output = new Rectangle(x - textOffsetX, y - textOffsetY, x, y);
+                break;
+            case LEFT:
+                // Text is positioned centered-left of x,y
+                textOffsetY /= 2;
+                output = new Rectangle(x - textOffsetX, y - textOffsetY, x, y + textOffsetY);
+                break;
+            case BELOW_RIGHT:
+                // Text is positioned below-right of x,y
+                output = new Rectangle(x, y, x + textOffsetX, y + textOffsetY);
+                break;
+            case ABOVE_RIGHT:
+                // Text is positioned above-right of x,y
+                output = new Rectangle(x, y - textOffsetY, x + textOffsetX, y);
+                break;
+            case RIGHT:
+                // Text is positioned centered-right of x,y
+                textOffsetY /= 2;
+                output = new Rectangle(x, y - textOffsetY, x + textOffsetX, y + textOffsetY);
+                break;
+        }
+
+        pointTextContainer.clashRect = output;
+        pointTextContainer.clashRotationDegrees = newRotation.degrees;
+
+        return output;
+    }
+
+    /**
+     * Mainly used for debugging purposes.
+     *
+     * @return Clash rectangle transformed back to label space
+     */
+    public static Rectangle getClashRectTransformed(PointTextContainer pointTextContainer, double originX, double originY, Rotation rotation) {
+        Rectangle output = null;
+
+        final Rectangle clashBounds = pointTextContainer.clashRect;
+
+        if (clashBounds != null) {
+            final Rotation myRotation = new Rotation(rotation.degrees, 0, 0);
+
+            // "Undo" clash rectangle offset
+            Rectangle rect1 = clashBounds.shift(new Point(-pointTextContainer.horizontalOffset, -pointTextContainer.verticalOffset));
+
+            // "Undo" clash rectangle rotation
+            Point lt = myRotation.rotate(rect1.left, rect1.top, false);
+            Point rb = myRotation.rotate(rect1.right, rect1.bottom, false);
+
+            // Imitate drawing clash rectangle as a label
+            Point lt2 = lt.offset(-originX, -originY);
+            Point rb2 = rb.offset(-originX, -originY);
+            Point lt3 = rotation.rotate(lt2, true);
+            Point rb3 = rotation.rotate(rb2, true);
+            Rectangle rect2 = new Rectangle(lt3.x, lt3.y, rb3.x, rb3.y);
+            output = rect2.shift(new Point(pointTextContainer.horizontalOffset, pointTextContainer.verticalOffset));
+        }
+
+        return output;
     }
 }
--- a/mapsforge-map-android/build.gradle
+++ b/mapsforge-map-android/build.gradle
@@ -1,7 +1,9 @@
-apply plugin: 'java-library'
+apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
 
 dependencies {
+    testImplementation "junit:junit:4.13.2"
+    testImplementation 'org.robolectric:robolectric:4.14.1'
     compileOnly files("${System.env.ANDROID_HOME}/platforms/android-${androidCompileSdk()}/android.jar")
     api project(":mapsforge-map-reader")
     api 'com.caverock:androidsvg:1.4'
@@ -9,8 +11,14 @@ dependencies {
 
 publishing {
     publications {
-        maven(MavenPublication) {
-            from components.java
+        release(MavenPublication) {
+            groupId = group()
+            artifactId = 'mapsforge-map-android'
+            version = getVersion()
+
+            afterEvaluate {
+                from components.release
+            }
         }
     }
 }
@@ -23,4 +31,15 @@ if (project.hasProperty("SONATYPE_USERNAME")) {
 
 if (System.getenv('ANDROID_HOME') == null) {
     throw new GradleException("Environment variable ANDROID_HOME needs to be set to SDK folder")
+}
+
+android {
+    namespace 'org.mapsforge.map.android'
+    compileSdk androidCompileSdk()
+
+    publishing {
+        singleVariant('release') {
+            withSourcesJar()
+        }
+    }
 }
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
@@ -3,6 +3,7 @@
  * Copyright 2014 Ludwig M Brinckmann
  * Copyright 2015-2020 devemux86
  * Copyright 2019 Matthew Egeler
+ * Copyright 2024 Sublimis / Urban Biker
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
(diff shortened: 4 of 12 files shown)


mapsforge-map-android/build.gradle
```diff
@@ -1,16 +1,24 @@
-apply plugin: 'java-library'
+apply plugin: 'com.android.library'
```
devemux86, line 1: @Sublimis We do not want such a conversion.

Android module contains only Java code, so it should remain a simple Java module.

Only the tests do not justify such a conversion.
If they are useful, they can exist in their own separate module.

Sublimis: No problem to revert. Is there any downside to using com.android.library for an Android library?
Testing for label overlap is very important because it's easy to mess things up (as evidenced by the longevity of the bug).

devemux86: @Sublimis If a library contains only Java code and not Android resources,
there is no real reason to have a more complicated Android aar library.

devemux86:
@Sublimis Please check the code, it does not compile.

---

May I ask what happened in the other work with the hill shading improvements?
We discussed splitting it into multiple simpler pull requests.

If you're working on other areas of the library, that's perfectly fine :+1: 
We definitely need improvements in all areas!

devemux86:
@Sublimis 

May I ask what happened in the other work with the hill shading improvements?
We discussed splitting it into multiple simpler pull requests.

If you're working on other areas of the library, that's perfectly fine. :+1: 
We definitely need improvements in all areas!

Sublimis:
> @Sublimis Please check the code, it does not compile.

Haven't reverted build.gradle yet, please advise: should we revert or fix this error? It says the following, so the fix seems simple:


> To create a baseline, run `gradlew updateLintBaseline` after adding the following to the module's build.gradle file:
> /home/runner/work/mapsforge/mapsforge/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidCanvas.java:155: Error: Call requires **API level 9 (current min is 1)**: java.lang.String#isEmpty [NewApi]
> 
> ```
>   android {
>       lint {
>           baseline = file("lint-baseline.xml")
>       }
>   }
> ```
> 
>   For more details, see https://developer.android.com/studio/write/lint#snapshot
>

Sublimis:
> May I ask what happened in the other work with the hill shading improvements? We discussed splitting it into multiple simpler pull requests.
> 
> If you're working on other areas of the library, that's perfectly fine. 👍 We definitely need improvements in all areas!

This requires a bit more work to disentangle our playfulness, we'll definitely do a PR or two for this as soon as it's ready! 🙂

Sublimis:
Sorry for the third commit, for some reason build.gradle wasn't reverted even though we pushed the original version along with the other files.

mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
```diff
@@ -25,8 +26,40 @@ public interface Paint {
 
     int getTextHeight(String text);
 
+    default int getTextHeight(final String text, boolean includePadding) {
```
devemux86, line 29: The `default` methods in the interface are only used on Android.
So they can be hosted in `AndroidPaint` and avoid loading the interface.

Also, this is a feature of Java 8 and we're trying to be as backwards compatible as possible.

mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
```diff
@@ -25,8 +26,40 @@ public interface Paint {
...
+        int retVal = 0;
```
devemux86, line 30: Set variable 2 times?

mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
```diff
@@ -25,8 +26,40 @@ public interface Paint {
...
+            retVal += getFontPadding();
+        }
+
+        return retVal;
+    }
+
     int getTextWidth(String text);
 
+    default int getTextWidth(String text, boolean includePadding) {
+        int retVal = 0;
```
devemux86, line 44: Set variable 2 times?

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
```diff
@@ -151,6 +156,90 @@ public int getTextWidth(String text) {
...
+        if (text != null && widthMax >= 0) {
+
+            final TextPaint myPaint;
+            if (this.paint instanceof TextPaint) {
+                myPaint = (TextPaint) this.paint;
+            } else {
+                myPaint = new TextPaint(this.paint);
+            }
+
+            final StaticLayout layout = new StaticLayout(text, myPaint, widthMax, Layout.Alignment.ALIGN_NORMAL, 1, 0, includePadding);
```
devemux86, line 176: Please add in the method:
```java
@SuppressWarnings("deprecation")
```

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -128,4 +126,156 @@ public String toString() {
         stringBuilder.append(this.text);
         return stringBuilder.toString();
     }
+
+    /**
+     * @return New rotated x,y position of the container relative to the origin point
+     */
+    protected Point getRotRelPosition(double originX, double originY, Rotation rotation) {
```
devemux86, line 133: Better use full name `getRotatedRelativePosition` for clear understanding.

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -128,4 +126,156 @@ public String toString() {
...
+        return getRotRelPosition(this, originX, originY, rotation);
+    }
+
+    /**
+     * @return New rotated x,y position of the container relative to the origin point
+     */
+    public static Point getRotRelPosition(PointTextContainer pointTextContainer, double originX, double originY, Rotation rotation) {
```
devemux86, line 140: Better use full name `getRotatedRelativePosition` for clear understanding.

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -128,4 +126,156 @@ public String toString() {
...
+    }
+
+    /**
+     * See org.mapsforge.map.rendertheme.renderinstruction.Caption#Caption,
+     * org.mapsforge.map.android.graphics.AndroidPointTextContainer#AndroidPointTextContainer,
+     * or org.mapsforge.map.awt.graphics.AwtPointTextContainer#AwtPointTextContainer.
+     *
+     * @return Clash rectangle in absolute coordinate space
+     */
+    public static Rectangle getClashRect(PointTextContainer pointTextContainer, Rotation rotation) {
```
devemux86, line 164: Better use full name `getClashRectangle`.

Sublimis:
> @Sublimis Aren't Android improvements possible on the desktop as well (e.g. padding, ...) ?

Sure, but we're talking only about font padding and drawing clash rectangles outline for debugging, that could probably be done easily, but it's not our primary platform so we left that for somebody else to do if possible.

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -128,4 +126,156 @@ public String toString() {
...
+
+        return output;
+    }
+
+    /**
+     * Mainly used for debugging purposes.
+     *
+     * @return Clash rectangle transformed back to label space
+     */
+    public static Rectangle getClashRectTransformed(PointTextContainer pointTextContainer, double originX, double originY, Rotation rotation) {
```
devemux86, line 255: Better use full name `getClashRectangleTransformed`.

devemux86:
> Sure, but we're talking only about font padding and drawing clash rectangles outline for debugging, that could probably be done easily, but it's not our primary platform so we left that for somebody else to do if possible.

Is it just for debugging? The screenshots show otherwise.
In any case both platforms are evolving simultaneously, we can't leave one behind.
In such a case the code should only exist in the platform-specific classes.

devemux86:
> Sure, but we're talking only about font padding and drawing clash rectangles outline for debugging, that could probably be done easily, but it's not our primary platform so we left that for somebody else to do if possible.

@Sublimis Is it just for debugging? The screenshots show otherwise.
In any case both platforms are evolving simultaneously, we can't leave one behind.
In such a case the code should only exist in the platform-specific classes.

Sublimis:
> @Sublimis Is it just for debugging? The screenshots show otherwise. In any case both platforms are evolving simultaneously, we can't leave one behind. In such a case the code should only exist in the platform-specific classes.

Yes, `PointTextContainer.DEBUG_CLASH_BOUNDS` flag is useful just for debugging labels overlap, why else would we draw rectangles around labels? 🙂

devemux86:
@Sublimis The description mentions:

> Android: Label space now includes a small padding when checking for overlap (avoids placing labels too close together, which can sometimes appear as one word)

Is this only for debugging?

mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
Sublimis: We can transfer those to implementation classes, but then AWT will be left without label font padding concept. Maybe better to leave signatures in the interface and implement in both classes.

Sublimis:
> @Sublimis The description mentions:
> 
> > Android: Label space now includes a small padding when checking for overlap (avoids placing labels too close together, which can sometimes appear as one word)
> 
> Is this only for debugging?

No, this is not only for debugging. This is (tiny) label padding we're talking about, it has nothing to do with the `PointTextContainer.DEBUG_CLASH_BOUNDS` mentioned above, sorry for the confusion.

mapsforge-core/src/main/java/org/mapsforge/core/graphics/Paint.java
devemux86: We will have to choose one of 2 alternatives,
either implement and use the code on both platforms
or put the code in the platform class that is only used.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
```diff
@@ -171,6 +185,45 @@ public int getTextWidth(String text) {
...
+
+        return getFontPadding(fontMetrics);
+    }
+
+    /**
+     * Returns total vertical font padding (top+bottom). The value can also be used for horizontal padding.
+     * The value is computed as 12.5% of {@code fontMetrics.getHeight()}.
+     */
+    public static int getFontPadding(FontMetrics fontMetrics) {
+        return (int) (0.125 * fontMetrics.getHeight());
```
devemux86, line 224: Desktop also has similar font metrics methods to use instead of fixed padding.
(see `AwtPointTextContainer`)

devemux86, line 224: @Sublimis Desktop also has similar font metrics methods to use instead of fixed padding.
(see `AwtPointTextContainer`)

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
```diff
@@ -146,11 +151,118 @@ public int getTextHeight(String text) {
...
+    }
+
+    /**
+     * Returns total vertical font padding (top+bottom). The value can also be used for horizontal padding.
+     * The value is computed as (ascent - top + bottom).
+     */
+    public static int getFontPadding(android.graphics.Paint paint) {
+        final android.graphics.Paint.FontMetricsInt fontMetricsInt = paint.getFontMetricsInt();
+
+        return fontMetricsInt.ascent - fontMetricsInt.top + fontMetricsInt.bottom;
```
devemux86, line 263: @Sublimis Please explain the padding calculation details.

So we can apply a similar calculation in the `AwtPaint`.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
Sublimis: > Desktop also has similar font metrics methods to use instead of fixed padding. (see `AwtPointTextContainer`)

Unfortunately, that is not true, Android supports font metrics that are not available on AWT (as far as we know). For the easiest explanation, see here. Android:

image


AWT:

img1

In AWT there's no TOP and BOTTOM that are used to calculate padding on Android. In any case, the exact amount of padding is not essential at all, as long as it's some well-defined small fraction of the letter size. It's only used to separate labels that are very close together. Frankly, it would probably be easier to use a fraction in Android as well. Also please note that such fraction padding cannot be considered a *fixed* padding, as it depends on letter size. As such, it serves its purpose very well (to separate labels).

Or are you concerned about the value and/or hard-coding the fraction constant, 0.125? This could be factored out to a field, or e.g. `Parameters` config class, or even given as an option to the user (although not a very useful option...).

devemux86: @Sublimis Thanks, I have seen this image online.

Why 12.5%, why not 10%?

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
```diff
@@ -171,6 +185,45 @@ public int getTextWidth(String text) {
...
+        if (includePadding && retVal > 0) {
+            retVal += getFontPadding(fontMetrics);
+        }
+
+        return retVal;
+    }
+
+    @Override
+    public int getTextWidth(final String text, final int widthMax, boolean includePadding) {
+        return getTextWidth(text, includePadding);
```
devemux86, line 204: Here we can use code from `AwtPointTextContainer` to measure the width instead of default.

Like we use in `AndroidPaint` code from `AndroidPointTextContainer`

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
Sublimis: > Why 12.5%, why not 10%?

This is a power-of-two number so has an exact binary representation, but that aside, it also has a more practical property for us. On standard screens (dot pitch values cca 0.25-0.27mm) and with AWT sample app defaults, the calculated padding with 10% constant is only 1 pixel for many labels, while with the 12.5% it results in a padding of 2 (our padding is defined as top+bottom, so this is appealing because each side gets 1 pixel of padding).

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
devemux86: Also it is better to use the `StaticLayout.Builder` for newer Androids.
See the implementation in `AndroidPointTextContainer`.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
devemux86: @Sublimis Thanks, please add this explanation as comment in the code.
Otherwise in the future it will be difficult to understand what happens.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
```diff
@@ -146,11 +151,118 @@ public int getTextHeight(String text) {
...
+    /**
+     * Uses StaticLayout to measure text accurately, and respects Spans.
+     * Returns width of the widest line.
+     */
+    @SuppressWarnings("deprecation")
+    @Override
+    public int getTextWidth(final String text, final int widthMax, boolean includePadding) {
+        int retVal = 0;
+
+        if (text != null && widthMax >= 0) {
```
devemux86, line 190: @Sublimis Why `widthMax >= 0` and not `widthMax > 0`?

devemux86:
@Sublimis Thanks, it looks very nice work! :+1: 

Please just check my comments on the code and if / how we can improve some things.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
Sublimis: This snippet is taken from our app so not sure any more, however by looking at the StaticLayout, width parameter value of zero is valid, so that is probably the reason. Although there's no explanation anywhere in the StaticLayout documentation why width=0 is accepted. But it is.

Note that these three-argument `getTextWidth(, ,)` methods are not currently used anywhere, but thought they might be useful in the future (the AndroidPaint one, that is). We used it a little while developing this PR, until decided on a different approach. They can be safely deleted for now (from the Paint interface too), you decide.

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
Sublimis: It's probably easier to just remove these three-argument getTextWidth methods, not used anyway.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
devemux86: > Note that these three-argument getTextWidth(, ,) methods are not currently used anywhere
> They can be safely deleted for now (from the Paint interface too), you decide.

If they are not used, it is better to remove them, as the public API is difficult to remove in the future.

And we avoid for now the desktop implementation. :slightly_smiling_face:

devemux86: @Sublimis 

> Note that these three-argument getTextWidth(, ,) methods are not currently used anywhere
> They can be safely deleted for now (from the Paint interface too), you decide.

If they are not used, it is better to remove them, as the public API is difficult to remove in the future.

And we avoid for now the desktop implementation. :slightly_smiling_face:

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
devemux86: > It's probably easier to just remove these three-argument getTextWidth methods, not used anyway.

@Sublimis Better do this. :+1:

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidPaint.java
Sublimis: Very true!

mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java
```diff
@@ -171,6 +186,45 @@ public int getTextWidth(String text) {
...
+    }
+
+    /**
+     * <p>
+     * Returns total vertical font padding (top+bottom). The value can also be used for horizontal padding.
+     * The value is computed as 12.5% of {@code fontMetrics.getHeight()}.
+     * <p>
+     * Why 12.5%, why not 10%?
+     * <p>
+     * 0.125 is a power-of-two number so has an exact binary representation, but that aside, it also has a more practical property for us. On standard screens (dot pitch values cca 0.25-0.27mm) and with AWT sample app defaults, the calculated padding with 10% constant is only 1 pixel for many labels, while with the 12.5% it results in a padding of 2 (our padding is defined as top+bottom, so this is appealing because each side gets 1 pixel of padding).
```
devemux86, line 222: Please split the comment in multiple lines here.

devemux86: @Sublimis Please split this comment in multiple lines.

Sublimis:
@devemux86 Sorry, it was multiline when pasted, then something happened in the meantime.

devemux86:
@Sublimis Is there anything else or I can merge it?

Sublimis:
> @Sublimis Is there anything else or I can merge it?

Just a sec.

Sublimis:
We now also have `drawClashBounds()` method for AWT (for debugging purposes)

devemux86:
> We now also have drawClashBounds() method for AWT (for debugging purposes)

@Sublimis This is a useful debugging function on both platforms. :+1:

Sublimis:
> @Sublimis Is there anything else or I can merge it?

We also have some label performance and overlapping improvements in Database/DirectRenderer and CanvasRasterer, but it's maybe better to do that as a separate PR, as it also needs some more profiling.

devemux86:
@Sublimis The `drawClashBounds` draws until the baseline, without the descent?

Sublimis:
> @Sublimis The `drawClashBounds` draws until the baseline, without the descent?

We can look into this if you're willing to wait a bit more

devemux86:
> We can look into this if you're willing to wait a bit more

@Sublimis There is no hurry, please take your time. :slightly_smiling_face:

devemux86:
@Sublimis What I mean is that during the clash comparison the examined rectangles contain the descent or not?

If they do not contain the descent then the current debug rectangles should be correct.

Sublimis:
It seems that they do not include descent, and it seems like being caused by the AwtPaint.getTextHeight() which was written by somebody else (our two-argument overload is just a copy). Can you look into fixing this method(s) to include descent?

devemux86:
@Sublimis Even if we change the `getTextHeight` and it returns a larger value,
the rectangle is still wrongly shifted to the top with bottom at baseline.

It seems more a painting problem?

Sublimis:
> @Sublimis Even if we change the `getTextHeight` and it returns a larger value, the rectangle is still wrongly shifted to the top with bottom at baseline.
> 
> It seems more a painting problem?

This is just a matter of including the label padding into calculations, we took care of it now

Sublimis:
However the descents are still popping out of the bottom, not a big deal, but improvement of AwtPaint.getTextHeight () methods is the only way to fix this completely.

devemux86:
@Sublimis What is needed? Change the line:
```java
return (int) this.font.createGlyphVector(fontMetrics.getFontRenderContext(), text).getVisualBounds().getHeight();
```

devemux86:
@Sublimis As mentioned above, even if we change the `getTextHeight`, even we make it to return 2x value,
the rectangle still starts at baseline.

Sublimis:
> @Sublimis What is needed? Change the line:
> 
> ```java
> return (int) this.font.createGlyphVector(fontMetrics.getFontRenderContext(), text).getVisualBounds().getHeight();
> ```

Yes, exactly that. Clash rectangle height is basically a text container boundary height, and container boundary is computed using that code.

Sublimis:
> @Sublimis As mentioned above, even if we change the `getTextHeight`, even we make it to return 2x value, the rectangle still starts at baseline.

We took care of that now, as said earlier 🙂

Sublimis:
> > @Sublimis As mentioned above, even if we change the `getTextHeight`, even we make it to return 2x value, the rectangle still starts at baseline.
> 
> We took care of that now, as said earlier 🙂

Here you can see an example with exaggerated padding:

image

devemux86:
@Sublimis Even on Android the rectangles seem attached at descent with much more free space at the top.
It appears on the first screenshots for "Gostingen". The padding should have same free space on all 4 sides?

Sublimis:
> @Sublimis Even on Android the rectangles seem attached at descent with much more free space at the top. It appears on the first screenshots for "Gostingen". The padding should have same free space on all 4 sides?

Ok, see it now. Please wait until we look into it.

devemux86:
@Sublimis More important is if the padding is used correctly in the label placement calculations?

The debug rectangles are less important, we can improve them if the calculations work correctly.

devemux86:
@Sublimis More important is if the padding is used correctly in the label placement calculations?
e.g. it is not shifted upwards like in the debug rectangles.

The debug rectangles are less important, we can improve them if the calculations work correctly.

devemux86:
> It seems like this cannot be fixed without knowing exactly where the baseline starts in the boundary/clash rectangle, which would require more coding

@Sublimis For the debug rectangles? So it is just a painting issue?

Sublimis:
> @Sublimis More important is if the padding is used correctly in the label placement calculations? e.g. it is not shifted upwards like in the debug rectangles.
> 
> The debug rectangles are less important, we can improve them if the calculations work correctly.

Please note these rectangles are used for labels clashing decisions. However yes, if some descent here or there touches another label is not so bad, probably.

Sublimis:
@devemux86 
That should be it.

devemux86:
@Sublimis So we are ready for merge now?

Sublimis:
Yes, unless you are willing to wait some more hours (or until tomorrow) so we maybe take descent value into consideration too 🙂 . The rough idea is (both Android and AWT): Add getDescent() to the Paint interface, implement, and use that to obtain descent value when drawing label. Move the y-coordinate up by this value before drawText call.

devemux86:
@Sublimis Sure we can wait and try if there is a better implementation. :+1:

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Caption.java
```diff
@@ -94,28 +95,6 @@ public Caption(GraphicFactory graphicFactory, DisplayModel displayModel, String
...
-                break;
-            case BELOW_RIGHT:
-            case ABOVE_RIGHT:
-            case RIGHT:
-                this.stroke.setTextAlign(Align.LEFT);
-                this.fill.setTextAlign(Align.LEFT);
-                break;
-            default:
-                throw new IllegalArgumentException("Position invalid");
-        }
```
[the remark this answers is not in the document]
Sublimis: Those lines interfere with correct drawing of labels, but only on Android, AWT lives just fine without it:

https://github.com/mapsforge/mapsforge/blob/698cf613f0c0694f0f385440d3ef9e03794f75f7/mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/graphics/AwtPaint.java#L247-L250

So removing label Paint-level align on Android basically makes the code that draws the labels more symmetrical and simpler. Do you think there are any drawbacks to this? Haven't found any.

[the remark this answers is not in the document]
devemux86: @Sublimis Tests must be made with the `position` parameter of the `caption` rule.
With `master` branch this section is needed, otherwise it breaks the Android rendering.

With the PR code it seems to not be needed. That's why I ask where it is handled now?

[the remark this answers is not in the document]
Sublimis: > @Sublimis Tests must be made with the `position` parameter of the `caption` rule. With `master` branch this section is needed, otherwise it breaks the Android rendering.
> 
> With the PR code it seems to not be needed. That's why I ask where it is handled now?

Here: https://github.com/mapsforge/mapsforge/pull/1579#discussion_r1858330986

devemux86:
@Sublimis It seems working fine now. Let me know if there is more work here or more testing is needed.

Sublimis:
That should be all regarding caption/point text labels for this PR. What did you mean by saying "tests must be made..."?

devemux86:
> What did you mean by saying "tests must be made..."?

@Sublimis That we should test all `position` values in `caption` theme rule on Android + Desktop,
to make sure everything continues to work correctly like before.

e.g. locate a restaurant on the map and change the `position` on its `caption` theme rule to all available values.

PR #1580, opened 2024-11-23
Desktop samples use label layer

devemux86:


Sublimis:
Let's transfer [this](https://github.com/mapsforge/mapsforge/commit/698cf613f0c0694f0f385440d3ef9e03794f75f7#r149654903) discussion here please.

In fact yes, you can reproduce by reducing the cache size, for example by changing this line: 

https://github.com/mapsforge/mapsforge/blob/869a7eeef84c99cadc270df9f1056bd7990ea8a2/mapsforge-map-awt/src/main/java/org/mapsforge/map/awt/util/AwtUtil.java#L41

to this:

`        TileCache firstLevelTileCache = new InMemoryTileCache(4);`

This seems to happen when some `MapElementContainer` that needs to be drawn has already been flushed from the cache (or maybe not inserted at all?).

Can this happen during normal operation? Relying on cache size seems like bad practice (if our theory is correct).

devemux86:
@Sublimis 
Cache in memory or storage is a requirement for map rendering on Android and desktop.
Calculated from screen resolution and tile size. Memory cache usually covers the visible tiles.
We cannot limit it and then expect the library to work properly.

Sublimis:
Here is the real reason (the above seems to be just a lucky coincidence).

It can be reproduced in master if the `DatabaseRenderer` constructor is called with `renderLabels=true` and `labelStore=null`. Of course, the sample app took care not to set the parameters this way, but that's not the point, the point is that this case is not mentioned in the constructor's javadoc, so it can be misleading.

Probably nothing important, so we can drop it, but in case anyone will be wondering.

The code:

https://github.com/mapsforge/mapsforge/blob/869a7eeef84c99cadc270df9f1056bd7990ea8a2/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java#L52-L79

devemux86:
@Sublimis Please note that there are 3 ways to render the map labels:
- Labels on map tiles (good if don't use map rotation)

- Label layer cached, see [LabelLayerUsingLabelCacheMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingLabelCacheMapViewer.java)

- Label layer threaded not cached, [LabelLayerUsingMapDataStoreMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingMapDataStoreMapViewer.java)

3rd case may seem preferable, but label rendering is slower than 2nd case with large complex maps.

devemux86:
@Sublimis Please note that there are 3 ways to render the map labels:
- Labels on map tiles (good if don't use map rotation)

- Label layer cached, see [LabelLayerUsingLabelCacheMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingLabelCacheMapViewer.java)

- Label layer (threaded) not cached with separate label store, see [LabelLayerUsingMapDataStoreMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingMapDataStoreMapViewer.java)

3rd case may seem preferable, but label rendering is slower than 2nd case with large complex maps.

Issue #319, opened 2014-11-06
Provide a low detail world map


mapsforge:
_From [h...@flashlight.de](https://code.google.com/u/105864613744213469970/) on May 27, 2012 22:20:10_

How about providing a low detail world map with small size?

It should contain all country borders plus their major cities.

If you do not want to provide that kind of map could you 
give some hint how to produce something like this in a
efficient way?

_Original issue: http://code.google.com/p/mapsforge/issues/detail?id=319_

andreynovikov:
Can you explain briefly how you managed to create such a map? I've tried to create my own one using only sea and land polygons and got 18MB file but yours is only 3MB.

devemux86:
@andreynovikov see also here:
https://groups.google.com/d/msg/mapsforge-dev/uaZM4NIydus/q4kOH7ibJtwJ

devemux86:
By the way, since we have now a low detail world map, I'm closing this issue.

devemux86:
The process for creating a low detail world map is explained here:
https://groups.google.com/forum/#!topic/mapsforge-dev/1PzOYvFs8sE

DrMarkEy:
I would like to use the world.map file with the major highways and borders. Unfortunately the attachments seem to be broken due to the migration from google code to github. Could someone provide me a link to this file?

devemux86:
The `world.map` containing land polygons is provided in our [server](http://download.mapsforge.org/maps/world/) with creation instructions [here](https://github.com/mapsforge/mapsforge/blob/master/docs/MapCreation.md#world-map).

gcPhil:
> The `world.map` containing land polygons is provided in our [server](http://download.mapsforge.org/maps/world/) with creation instructions [here](https://github.com/mapsforge/mapsforge/blob/master/docs/MapCreation.md#world-map).

What [DrMarkEy](https://github.com/DrMarkEy) was searching is the `world.map` with land polygons, borders and major Highways from [this post](http://code.google.com/p/mapsforge/issues/detail?id=319#hc6). Unfortunately the attachment seem to be broken due to the migration from google code to github.

PR #1582, opened 2024-11-26
Multi-map improvements

Sublimis:
~~* MultiMapDataStore.DataPolicy.RETURN_AUTOFILL: Return data from all datasets in the sequence that support a tile, up to the one that contains data for the entire tile (usually a world map). In other words, once a dataset fills the entire tile, the remaining datasets are not processed for that tile.~~

Here we are solving the following multi-map artifacts:

1. Low-res map coastline peeks out from behind high-res map (RETURN_ALL issue)

2. Incomplete tile returned on map edge (RETURN_FIRST issue; RETURN_ALL has this issue only when zoom levels are limited as in the sample app)

3. Return data from all maps that overlap on a tile

Motivation: Prevent ugly gaps at the edges of the map when the map does not cover the tile completely, and we have a background world map that should easily cover this gap. Additionally, prevent the land areas of the (low-res) world map from "peeking" behind the land areas of a (hi-res) main map, when the main map can provide the entire tile itself.

These problems can be easily seen in the Android application sample MultiMapLowResWorld.

### DIFF
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
@@ -238,6 +238,12 @@ public class MapFile extends MapDataStore {
             this.databaseIndexCache = new IndexCache(this.inputChannel, INDEX_CACHE_SIZE);
 
             this.timestamp = mapFile.lastModified();
+
+            // 2024: This may be removed in the future when world.map users get used to the map priorities feature
+            if ("world.map".equalsIgnoreCase(mapFile.getName())) {
+                // Some random (very) negative value no one will use
+                setPriority(Integer.MIN_VALUE + 374);
+            }
         } catch (Exception e) {
             // make sure that the channel is closed
             closeFileChannel();
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
@@ -238,12 +238,6 @@ public class MapFile extends MapDataStore {
             this.databaseIndexCache = new IndexCache(this.inputChannel, INDEX_CACHE_SIZE);
 
             this.timestamp = mapFile.lastModified();
-
-            // 2024: This may be removed in the future when world.map users get used to the map priorities feature
-            if ("world.map".equalsIgnoreCase(mapFile.getName())) {
-                // Some random (very) negative value no one will use
-                setPriority(Integer.MIN_VALUE + 374);
-            }
         } catch (Exception e) {
             // make sure that the channel is closed
             closeFileChannel();
--- a/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
+++ b/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
@@ -6,6 +6,7 @@
  * Copyright 2016 bvgastel
  * Copyright 2017 linuskr
  * Copyright 2017 Gustl22
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -1073,9 +1074,27 @@ public class MapFile extends MapDataStore {
     }
 
     @Override
+    public boolean supportsFullTile(Tile tile) {
+        return supportsFullArea(tile.getBoundingBox(), tile.zoomLevel);
+    }
+
+    @Override
     public boolean supportsArea(BoundingBox boundingBox, byte zoomLevel) {
         return boundingBox.intersects(getMapFileInfo().boundingBox)
                 && (zoomLevel >= this.zoomLevelMin && zoomLevel <= this.zoomLevelMax);
+    }
+
+    @Override
+    public boolean supportsFullArea(BoundingBox boundingBox, byte zoomLevel) {
+        final BoundingBox bbox1 = getMapFileInfo().boundingBox;
+        final BoundingBox bbox2 = boundingBox;
+        return bbox1.intersects(bbox2)
+                && zoomLevel >= this.zoomLevelMin
+                && zoomLevel <= this.zoomLevelMax
+                && bbox1.contains(bbox2.maxLatitude, bbox2.maxLongitude)
+                && bbox1.contains(bbox2.minLatitude, bbox2.minLongitude)
+                && bbox1.contains(bbox2.maxLatitude, bbox2.minLongitude)
+                && bbox1.contains(bbox2.minLatitude, bbox2.maxLongitude);
     }
 
     /**
--- a/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
@@ -1,6 +1,7 @@
 /*
  * Copyright 2014-2015 Ludwig M Brinckmann
  * Copyright 2015-2022 devemux86
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -25,11 +26,18 @@ import java.util.List;
 /**
  * A MapDatabase that reads and combines data from multiple map files.
  * The MultiMapDatabase supports the following modes for reading from multiple files:
+ * <p>
  * - RETURN_FIRST: the data from the first database to support a tile will be returned. This is the
  * fastest operation suitable when you know there is no overlap between map files.
+ * <p>
+ * - RETURN_AUTOFILL: Return data from all datasets in the sequence that support a tile, up to the one
+ * that contains data for the entire tile (usually a world map). In other words, once a dataset fills
+ * the entire tile, the remaining datasets are not processed for that tile.
+ * <p>
  * - RETURN_ALL: the data from all files will be returned, the data will be combined. This is suitable
  * if more than one file can contain data for a tile, but you know there is no semantic overlap, e.g.
  * one file contains contour lines, another road data.
+ * <p>
  * - DEDUPLICATE: the data from all files will be returned but duplicates will be eliminated. This is
  * suitable when multiple maps cover the different areas, but there is some overlap at boundaries. This
  * is the most expensive operation and often it is actually faster to double paint objects.
@@ -38,6 +46,7 @@ public class MultiMapDataStore extends MapDataStore {
 
     public enum DataPolicy {
         RETURN_FIRST, // return the first set of data
+        RETURN_AUTOFILL, // return data until one dataset fills the entire tile
         RETURN_ALL, // return all data from databases
         DEDUPLICATE // return all data but eliminate duplicates
     }
@@ -109,6 +118,7 @@ public class MultiMapDataStore extends MapDataStore {
                     }
                 }
                 return 0;
+            case RETURN_AUTOFILL:
             case RETURN_ALL:
             case DEDUPLICATE:
                 long result = 0;
@@ -132,16 +142,18 @@ public class MultiMapDataStore extends MapDataStore {
                     }
                 }
                 return null;
+            case RETURN_AUTOFILL:
+                return readLabels(tile, false, true);
             case RETURN_ALL:
-                return readLabels(tile, false);
+                return readLabels(tile, false, false);
             case DEDUPLICATE:
-                return readLabels(tile, true);
+                return readLabels(tile, true, false);
         }
         throw new IllegalStateException("Invalid data policy for multi map database");
 
     }
 
-    private MapReadResult readLabels(Tile tile, boolean deduplicate) {
+    private MapReadResult readLabels(Tile tile, boolean deduplicate, boolean sequential) {
         MapReadResult mapReadResult = new MapReadResult();
         for (MapDataStore mdb : mapDatabases) {
             if (mdb.supportsTile(tile)) {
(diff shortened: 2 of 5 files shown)


devemux86:
@Sublimis Can you show screenshots with the difference?

> once a dataset fills the entire tile, the remaining datasets are not processed for that tile

Note that world map supports all tiles BUT does not have all or detailed data.

At least until someone provides a world map with more data:
- https://github.com/mapsforge/mapsforge/issues/319#issuecomment-61953086
- https://github.com/devemux86/cruiser/discussions/101

So the small maps with detailed data are more important in rendering.
World map is only useful at small zoom levels as shown in `MultiMapLowResWorld`.

Sublimis:
> Note that world map supports all tiles BUT does not have all or detailed data.
> 
> At least until someone provides a world map with more data:
> 
>     * [Provide a low detail world map #319 (comment)](https://github.com/mapsforge/mapsforge/issues/319#issuecomment-61953086)
> 
>     * [World map devemux86/cruiser#101](https://github.com/devemux86/cruiser/discussions/101)
> 
> 
> So the small maps with detailed data are more important in rendering. World map is only useful at small zoom levels as shown in `MultiMapLowResWorld`.

Note, when map files are added by priority, world map will be the last to be processed, and only if none of the previous maps provided an entire tile:

https://github.com/mapsforge/mapsforge/blob/f59ea94955f5e0de0fc28c1b9aa9e431c7f850b1/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MultiMapLowResWorld.java#L81-L85

devemux86:
It is still not clear what is the difference between before vs after.
For example we see a different map with `master` (not rectangle):

devemux86:
@Sublimis I edited your post with the screenshots to put them together for easy reading.
You check it and see the `img` rule that is useful to avoid filling the posts with large images. :slightly_smiling_face: 
```html
<img src="..." width="...%"/>
```

Sublimis:
> It is still not clear what is the difference between before vs after. For example we see a different map with `master` (not rectangle):
> 

Zoom in a bit more (twice) on the edge of the map, until scale shows "5 km". Our map was taken from here, but the exact map shouldn't matter much: https://www.openandromaps.org/en/downloads/europe

devemux86:
@Sublimis Is is about the rectangle at OAM map edges?
Mapsforge maps do not have it. Where does it come from?

Note: it is easier to test with a small map (with see) like Azores.

devemux86:
@Sublimis Is is about the rectangle at OAM map edges?
Mapsforge maps do not have it.
Where does it come from? Is it in the map data?

Note: it is easier to test with a small map (with sea) like Azores.

Sublimis:
Map source doesn't matter, as long as you find zoom level >10 and an edge with incomplete tile. Ok, we can try Azores or anything else.

devemux86:
@Sublimis Try to also reproduce it with Mapsforge maps.

Sublimis:
No problem 🙂  Azores map from mapsforge server:

| 20241126-azores-before1 | 20241126-azores-before2 |
| --------------------------------------- | --------------------------------------- |

Sublimis:
> No problem 🙂 Azores map from mapsforge server:

The second image just shows the zoom center!

Sublimis:
With this PR applied:


| 20241126-azores-after3 |  |
| --------------------------------------- | --------------------------------------- |

devemux86:
@Sublimis Thanks, I understand now, I was looking at different things! :slightly_smiling_face:

devemux86:
@Sublimis 

- Does the order of opened map files matter?
As we shouldn't expect apps to open multiple map files in a specific order.

- `RETURN_AUTOFILL` can be bit confusing as name (especially "auto").
Is there a better name?

- Can you also add this flag in desktop Samples?

devemux86:
@Sublimis What happens with adjacent maps where each map has a part of the data of their shared tile?
Does `break` only return the data of one map?

devemux86:
@Sublimis 

> Does the order of opened map files matter?
As we shouldn't expect apps to open multiple map files in a specific order.

The order of opened maps matters, if we add the world map first, then it covers everything.

> What happens with adjacent maps where each map has a part of the data of their shared tile?
Does `break` only return the data of one map?

Indeed, this does not seem to work properly.

For example, opening the maps of Austria and Switzerland in this order.
With `AUTOFILL` their shared tile is filled with only Austria data compared to `ALL`:

|AUTOFILL|
|-|
|AUTOFILL|

|ALL|
|-|
|ALL|

Sublimis:
>     * Does the order of opened map files matter?

Yes, maps should be inserted in order of precedence for this to function as advertised. Or at the very least, the world map (if used) should be inserted last. Anyone using the RETURN_FIRST is already doing this. If a user who *consciously chooses to use the RETURN_AUTOFILL flag* sees artifacts like the above, they will simply change the order of the inserted maps. (This should be put as a note in the description of the flag, if not already clear enough).

>       As we shouldn't expect apps to open multiple map files in a specific order.

This is incorrect. Anyone who uses the RETURN_FIRST flag already pays attention to the order of the maps inserted if there is a chance that more than one map can serve a particular tile, which is often the case. Especially if they are using the world map, which overlaps all the other maps.

>     * `RETURN_AUTOFILL` can be bit confusing as name (especially "auto").
>       Is there a better name?

Since the library decides ("automatically") when a tile is "filled", the name seems appropriate. We don't have a better name, if anyone can think of a better name, please suggest it!

>     * Can you also add this flag in desktop Samples?

Yes

Sublimis:
> The order of opened maps matters, if we add the world map first, then it covers everything.
> 
> > What happens with adjacent maps where each map has a part of the data of their shared tile?
> > Does `break` only return the data of one map?
> 
> Indeed, this does not seem to work properly.

It is working exactly as advertised. Note, this flag is not a replacement for RETURN_ALL!

devemux86:
@Sublimis Indeed `RETURN_FIRST` is not really very useful, as it produces half filled multi maps.
Why open the maps of Austria and Switzerland and see only Austria on their borders?

Also this new flag seems to have a very specific use, like world map + 1 country on top?
As seen above multiple countries will not work properly.

Please let's revert the samples to `master` code, as `ALL` is what most users would use.

Sublimis:
For your example above to work "correctly" with the AUTOFILL flag (i.e. to return data from both maps), neither map should claim to have data for an entire tile that it doesn't actually have. But that then becomes "just" a question of implementation details, i.e. how to get the exact map boundaries (currently this is a rough rectangle). That's why this flag is currently most useful with a world map.

devemux86:
@Sublimis We also need to think about the performance of the map.
Heavy calculations with thousands of map data on shared tiles can have negative results.

Sublimis:
> @Sublimis Indeed `RETURN_FIRST` is not really very useful, as it produces half filled multi maps. Why open the maps of Austria and Switzerland and see only Austria on their borders?

It seems you should think this through some more. Do take your time. RETURN_FIRST is useful when the user knows the maps they throw into the `MultiMapDataStore` are disjoint, i.e. not overlapping anywhere (if not using the world map). If they use the world map, this will also work, as the user will see their country as expected, but when they go to entirely different part of the world they will see the world map.

Sublimis:
Also note that no one is forcing anyone to use the new flag. It has its own benefit that no other flag covers. They all have artifacts in a common useful scenario as shown above.

The new flag is in addition to the current flags and does not change their current behavior in any way.

The sample application can be reverted to RETURN_ALL, no problem. However, there is no performance difference between the two flags for the sample application.

Sublimis:
> @Sublimis We also need to think about the performance of the map. Heavy calculations with thousands of map data on shared tiles can have negative results.

This doesn't really have anything to do with the new flag, does it? It may only be faster than RETURN_ALL, or exactly the same. The new flag can never be slower than RETURN_ALL.

devemux86:
@Sublimis I meant that users usually open maps of neighboring countries.
Of course there could be other use cases of the map library. :slightly_smiling_face:

Sublimis:
By the way, we first tried to implement such functionality at the "layer level", i.e. use RETURN_ALL (without limiting the zoom level of the world map), but somehow force the sea layer of the more important map to be drawn on top of the less important (world) map, to prevent coastline artifacts. However, we learned that ALL sea layers are hard-coded as layer zero, so no go. Maybe this is a part of the library that could also be improved in the future, but we will almost certainly not deal with it.

We also first used the name RETURN_SEQUENTIAL for the new flag, but this name doesn't really say much. That's why there are some `boolean sequential` parameters in the code, in case you were wondering. We should probably rename that.

The RETURN_AUTOFILL flag is perhaps best described as a "fixed" version of RETURN_FIRST; it behaves the same unless the tile is incomplete (partially drawn), in which case it will try to load more data (unlike RETURN_FIRST).

Also, our opinion is that no one should ever look at a map that doesn't have a world background when zoomed out, especially considering that the current world map is so lightweight. But that's just us 🙂

Sublimis:
Let's recap and propose a slightly different, more universal approach.

Here we are solving the following multi-map artifacts:

1. Low-res map coastline peeks out from behind high-res map (RETURN_ALL issue)

2. Incomplete tile returned on map edge (RETURN_FIRST issue; RETURN_ALL has this issue only when zoom levels are limited as in the sample app)

Now we also want to satisfy this useful condition:

3. Return data from all maps that overlap on a tile

The proposal is to provide the ability to assign priority to the inserted map. By default, this can be 0 for all maps via method overloading (so that no code changes are required from the user). Then return all data (just like RETURN_ALL would do), but sorted by priority. If a map has priority <0, return its data only if the tile is not yet filled by other maps. We can call this a background map priority.

Such functionality can even be implemented without introducing a new flag. By default it will behave just like the RETURN_ALL is behaving right now.

However, we could also introduce a small auxiliary functionality so that users of the RETURN_ALL + world map can immediately benefit from removing the "peeking coastline" artifact: The library could detect when a map is called "world.map" and automatically assign it a negative priority.

What do you think?

devemux86:
@Sublimis Thanks for the detailed description.

A universal solution seems better, so opening **azores** + **austria** + **switzerland** + **world**
would work correctly in all cases:
- Coastline of Azores
- Common tiles of Austria and Switzerland

devemux86:
@Sublimis It is also important if we manage not to break how the public API works.
As it is already used in many applications.

devemux86:
@Sublimis Can you change the title and the description of the pull request?

devemux86:
@Sublimis It does not work if we don't set priority:
```java
Caused by: java.lang.IllegalArgumentException: Duplicate map database
                 	at org.mapsforge.map.datastore.MultiMapDataStore.addMapDataStore(MultiMapDataStore.java:70)
                 	at org.mapsforge.samples.android.MultiMapLowResWorld.onCreate(MultiMapLowResWorld.java:86)
```

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
```diff
@@ -44,13 +48,13 @@ public enum DataPolicy {
...
     private final DataPolicy dataPolicy;
-    private final List<MapDataStore> mapDatabases;
+    private final Set<MapDataStore> mapDatabases;
     private LatLong startPosition;
     private byte startZoomLevel;
 
     public MultiMapDataStore(DataPolicy dataPolicy) {
         this.dataPolicy = dataPolicy;
-        this.mapDatabases = new ArrayList<>();
+        this.mapDatabases = new TreeSet<>();
```
devemux86, line 57: @Sublimis Why this change?

If the order of insertion is still important, we can sort the list in the app or in the library.

Sublimis:
There, that should do it.


image

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
Sublimis: Done that. Collections.sort() is guaranteed to be stable, so it will not change the insertion order if all map files happen to have the same priority (which is the default, so it retains the current behavior).

Sublimis:
> @Sublimis It does not work if we don't set priority. Also desktop samples does not work.
> 
> ```java
> java.lang.IllegalArgumentException: Duplicate map database
>     at org.mapsforge.map.datastore.MultiMapDataStore.addMapDataStore(MultiMapDataStore.java:70)
>     at org.mapsforge.samples.android.MultiMapLowResWorld.onCreate(MultiMapLowResWorld.java:86)
> ```

Fixed that, please try now with the latest code.

mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java
```diff
@@ -237,6 +238,12 @@ public MapFile(File mapFile, String language) {
             this.databaseIndexCache = new IndexCache(this.inputChannel, INDEX_CACHE_SIZE);
 
             this.timestamp = mapFile.lastModified();
+
+            // 2024: This may be removed in the future when world.map users get used to the map priorities feature
+            if ("world.map".equalsIgnoreCase(mapFile.getName())) {
+                // Some random (very) negative value no one will use
+                setPriority(Integer.MIN_VALUE + 374);
+            }
```
devemux86, line 246: @Sublimis We should not do this inside the library.
This must be done by the apps that know the map content.

Also, the map names are not fixed, they can be changed freely.
For example [paws.cz](https://osm.paws.cz/) has europe_admin map.

devemux86: @Sublimis Let's put it only in the sample apps.
Android has it, we can also put it in desktop samples.

devemux86: @Sublimis We can set it manually on desktop samples as well, just like on Android.
They are just examples.

Sublimis: It's actually better like this, without assuming that the world.map file should be present, so the user decides which files to use. Because desktop sample has no startup menu like the Android app, it's just a single app.

devemux86:
@Sublimis Now it seems to work ok in all cases.
If it is complete, we can merge it?

Sublimis:
Wait, maybe to clarify comments a bit.

Sublimis:
Done.

PR #1584, opened 2024-12-02
Proper implementation of MapElementContainer.compareTo(other)

Sublimis:
* Proper implementation of MapElementContainer.compareTo(other), consistent with equals, and which does not allow two elements to be arbitrarily ordered (as this makes the drawing of the elements non-deterministic)

* Tests for MapElementContainer.compareTo(other)


devemux86:
@Sublimis Thanks!

PR #1586, opened 2024-12-03
Completing deterministic labels and symbols effort

Sublimis:
1) Important improvements to `MapElementContainer.compareTo()` and `MapElementContainer.clashesWith()`

2) Important improvements to `PointTextContainer.clashesWith()`

3) Improvements to `LayerUtil.collisionFreeOrdered()`

The above improvements together make non-deterministic labels and symbols a history!

Additional:

* Prevent labels and symbols overlap when the overlapping items are both marked as "display always"

* `DirectRenderer` is now obsolete. Now that the labels are deterministic, it doesn't bring anything new to the table that isn't already covered by the `DatabaseRenderer`. It is thus converted into a simple "empty shell" subclass of the `DatabaseRenderer` with the same interface as before, so it won't break any apps that use it. (#1085)

* Slight performance improvement when drawing labels: Avoid sorting when not needed; Reverse order of a sorted list instead of sorting again; Use List-s instead of Set-s where possible.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
@@ -73,11 +73,7 @@ public class DatabaseRenderer extends StandardRenderer {
         this.tileCache = tileCache;
         this.labelStore = labelStore;
         this.renderLabels = renderLabels;
-        if (!renderLabels) {
-            this.tileDependencies = null;
-        } else {
-            this.tileDependencies = new TileDependencies();
-        }
+        this.tileDependencies = new TileDependencies();
     }
 
     /**
@@ -86,6 +82,7 @@ public class DatabaseRenderer extends StandardRenderer {
      * @param rendererJob the job that should be executed.
      */
     public TileBitmap executeJob(RendererJob rendererJob) {
+        TileBitmap output = null;
         RenderContext renderContext = null;
         try {
             renderContext = new RenderContext(rendererJob, new CanvasRasterer(graphicFactory));
@@ -126,28 +123,33 @@ public class DatabaseRenderer extends StandardRenderer {
                         renderContext.canvasRasterer.fillOutsideAreas(Color.TRANSPARENT, insideArea);
                     }
                 }
-                return bitmap;
+
+                output = bitmap;
+            } else {
+                // outside of map area with background defined:
+                output = createBackgroundBitmap(renderContext);
             }
-            // outside of map area with background defined:
-            return createBackgroundBitmap(renderContext);
         } catch (Exception e) {
             LOGGER.warning(e.toString());
-            return null;
         } finally {
             if (renderContext != null) {
                 renderContext.destroy();
             }
+            if (!rendererJob.labelsOnly && output != null) {
+                tileCache.put(rendererJob, output);
+                removeTileInProgress(rendererJob.tile);
+            }
         }
+
+        return output;
     }
 
     public MapDataStore getMapDatabase() {
         return this.mapDataStore;
     }
 
-    void removeTileInProgress(Tile tile) {
-        if (this.tileDependencies != null) {
-            this.tileDependencies.removeTileInProgress(tile);
-        }
+    protected void removeTileInProgress(Tile tile) {
+        this.tileDependencies.removeTileInProgress(tile);
     }
 
     /**
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/MapWorkerPool.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/MapWorkerPool.java
@@ -3,6 +3,7 @@
  * Copyright 2014 Ludwig M Brinckmann
  * Copyright 2015-2017 devemux86
  * Copyright 2016 ksaihtam
+ * Copyright 2024 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -151,10 +152,6 @@ public class MapWorkerPool implements Runnable {
                     return;
                 }
 
-                if (!rendererJob.labelsOnly && bitmap != null) {
-                    MapWorkerPool.this.tileCache.put(rendererJob, bitmap);
-                    MapWorkerPool.this.databaseRenderer.removeTileInProgress(rendererJob.tile);
-                }
                 MapWorkerPool.this.layer.requestRedraw();
 
                 if (DEBUG_TIMING) {


Sublimis:
Sorry for the additional commits, it was just about including the PointTextContainer which was forgotten (but is largely related to this PR).

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -64,23 +64,23 @@ protected PointTextContainer(Point point, double horizontalOffset, double vertic
 
     @Override
     public boolean clashesWith(MapElementContainer other, Rotation rotation) {
-        if (super.clashesWith(other, rotation)) {
-            return true;
```
devemux86, line 68: Isn't `super.clashesWith` needed anymore?

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
```diff
@@ -163,83 +163,96 @@ private TileBitmap createBackgroundBitmap(RenderContext renderContext) {
...
-                    // set of tilesInProgress inside the TileDependencies.
-                    labelsToDraw.addAll(tileDependencies.getOverlappingElements(neighbour, renderContext.rendererJob.tile));
-
-                    // but we need to remove the labels for this tile that overlap onto a tile that has been drawn
-                    for (MapElementContainer current : renderContext.labels) {
-                        if (current.intersects(neighbour.getBoundaryAbsolute())) {
-                            undrawableElements.add(current);
+    private void processLabels(RenderContext renderContext) {
+        final List<MapElementContainer> labelsToDraw;
+        {
```
devemux86, line 170: We can avoid this `{` so commit history is more clear with the changes.

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
Sublimis: > Isn't `super.clashesWith` needed anymore?

No, `super.clashesWith` does not take rotation into account, so it is not relevant for labels. The important part of the `super.clashesWith` was put here, i.e. the `Display.ALWAYS == this.display` checks.

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
Sublimis: Those are scope-reducing braces, so it's clear that the purpose of the rest of the code is to set the variable. In our opinion it is better to have more error-resistant code than not to have it. No problem to remove those, are you sure?

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DirectRenderer.java
```diff
@@ -125,7 +126,8 @@ public TileBitmap executeJob(RendererJob rendererJob) {
         }
     }
 
-    private Set<MapElementContainer> processLabels(RenderContext renderContext) {
+    /** TODO (2024): If this renderer is ever going to be used, please see how labels are handled in the {@link DatabaseRenderer}, and then implement here. */
```
devemux86, line 129: `DirectRenderer` is not used by Mapsforge.

It could be used by apps to get bitmap tiles from vector maps in their own custom map engines.
However, it has worse map rendering than Mapsforge map engine, especially with the label layer.

It would be better if all developers work and improve the Mapsforge map engine,
so that our resources are not separated and we all benefit from the improvements.

devemux86:
@Sublimis Is there still work to be done?

Sublimis:
Mostly done, but these upcoming simple yet far-reaching changes (which we're quite proud of) might cause some head scratching, so now we're investing time into detailed explanations. 🙂  As soon as those are finished, we'll send a new commit.

For example: The library will now not allow two labels *or symbols* that are **both** marked as `Display.ALWAYS` to overlap. If you feel this is wrong, we can discuss it. However, it makes the map less cluttered (e.g. those prominent gas-station symbols that cover the map over urban areas). We'll send some screenshots to explain our point. Of course, when only one of those items is marked as `Display.ALWAYS`, it will properly be displayed without overlapping with items **not** marked as `Display.ALWAYS`.

Also: `DirectRenderer` is now obsolete. Now that the labels are deterministic, it doesn't bring anything new to the table that isn't already covered by the `DatabaseRenderer`. It is thus converted into a simple "empty shell" subclass of the `DatabaseRenderer`, so it won't break any apps that use it.

#1085 
@monsieurtanuki

devemux86:
@Sublimis 

> The library will now not allow two labels or symbols that are both marked as `Display.ALWAYS` to overlap.

Labels / symbols marked with `Display.ALWAYS` must be displayed always, so they should overlap with anything.
Otherwise the "always" loses its meaning. If someone wants less labels / icons can remove the "always".

devemux86:
@Sublimis 

If you think there are improvements to the Motorider map theme, please write here:
- https://github.com/devemux86/cruiser/discussions/612

It is written by @MotoUKRider and we contribute it to the Mapsforge community.

devemux86:
@Sublimis 

> The library will now not allow two labels or symbols that are both marked as `Display.ALWAYS` to overlap.

> Of course, when only one of those items is marked as Display.ALWAYS, it will properly be displayed without overlapping with items not marked as Display.ALWAYS.

We do not want these, as it changes how "always" was designed (see [here](https://github.com/mapsforge/mapsforge/blob/master/docs/Rendertheme.md#display)).

Labels / symbols marked with `Display.ALWAYS` must be displayed always.
So they should overlap with anything (other element with or not"always").

If someone wants less labels / icons on the map can remove the "always".

Sublimis:
> Labels / symbols marked with `Display.ALWAYS` must be displayed always, so they should overlap with anything. Otherwise the "always" loses its meaning. If someone wants less labels / icons can remove the "always".

Can you please explain under what circumstances it is a good (or positive) thing to have two `Display.ALWAYS` items to overlap? Maybe we're missing something. Note, the items are still drawn in order of their priority; The `Display.ALWAYS` is not the only property taken into account when drawing items.

devemux86:
@Sublimis 

> Can you please explain under what circumstances it is a good (or positive) thing to have two Display.ALWAYS items to overlap?

Its name indicates its meaning: _"always: an element will always be displayed regardless of space"_.

Sublimis:
Here are some screenshots, so it's more clear.

## Before

2024-12-04-always_overlap_on


## After

2024-12-04-always_overlap_off-1
2024-12-04-always_overlap_off-2
2024-12-04-always_overlap_off-3
2024-12-04-always_overlap_off-4

Sublimis:
> Its name indicates its meaning: _"always: an element will always be displayed regardless of space"_.

In other words, ugly overlap of two "always" items should has less priority over the "always" property? The logic is there, but maybe someone should communicate this to the theme creators and take their opinion into account. We think the "display always" property as currently defined (and abused) is too powerful. After all, we're talking about labels and symbols here, not rocket science...

In any case, this is really just a minor change in the code, easily reverted (and un-reverted in the future if needed).

Sublimis:
> By the way the "always" is also used in oneway arrows. Have you checked if fewer oneway arrows now appear?

Will do this now and get back to you!

Sublimis:
> > By the way the "always" is also used in oneway arrows. Have you checked if fewer oneway arrows now appear?
> 
> Will do this now and get back to you!

Did you mean something like this?

2024-12-04-always_overlap_off-arrows

devemux86:
@Sublimis 

> Did you mean something like this?

That's it, can you check the oneway arrows before and after the change?

Sublimis:
One more thing to point out. The `Display.ALWAYS` issue in question does not affect the determinism(*) of the symbols, it's just a matter of how the map looks 👍  So, it's really not that big of a deal anyway.

(*) Maybe you can notice how gas station symbols in the first (overlapping) screenshot above are nicely drawn from left to right 💯 🙂

devemux86:
@MotoUKRider 

It's more about the same elements with "always" rule. Please the screenshots with fuel stations above:
- https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-2517075516

devemux86:
@Sublimis @MotoUKRider 

We can probably merge it and see it in action within the apps.

Note: Mapsforge samples don't have layers menu, they display everything in the map theme.

Sublimis:
Regarding the one-way arrows. They are generally nicely spaced, so do not overlap, and the before/after is usually the same. Here are screenshots of one instance where we did found overlap and a before/after difference. In our opinion it's a minor difference, @MotoUKRider what do you think?

## Before

2024-12-04-always_overlap_off-arrows-2-v0 22 0


## After

2024-12-04-always_overlap_off-arrows-2

Sublimis:
@MotoUKRider 

> I now see what you mean about the Fuel icons. So, essentially, why show 10 clustered together when a single will work (at that zoom level) and then allow them to appear as the user zooms in? There's a similar approach to "grouping" with "Favorite" waypoints in Cruiser (ie: a single object with a "5" caption and once you zoom in a bit more, the individual icons are displayed - I guess that's the jist of this thread.

Exactly that, the idea was to avoid as much symbol overlapping as possible which just causes clutter and benefits almost nothing, but natively, without including complications like the cluster feature (which would also handle things differently and display a different, maybe unrelated, icon). Is there anything in the render theme that could help accomplish this, or is it okay the way it is done now (only in the code)?

devemux86:
@MotoUKRider If `display` is not set, then the `ifspace` is the default.

Sublimis:
Here is the commit.

devemux86:
@Sublimis Is this ready to merge?

devemux86:
@Sublimis Is this ready to merge?

Maybe you can update the pull request description with the details of the new commits?

Sublimis:
> @Sublimis Is this ready to merge?

A few more details will be added, sorry.

1) One more change regarding DatabaseRenderer/MapWorkerPool (without changing behavior) that was planned for later should probably be included here, because it's closely related to what was mentioned at #1085 and because we are changing the DirectRenderer in this PR:

>     * in this last case, potentially because of a bug - as elements are never removed from `tileDependencies` _except_ in `MapWorkerPool`, that we don't want to use because of the display refresh

2) One more slight optimization (no behavior change) in `LayerUtil.collisionFreeOrdered()`

> Maybe you can update the pull request description with the details of the new commits?

Will do.

By the way, it appears that symbol clash check does not take rotation into account, maybe something that should be considered in the future.

Sublimis:
@devemux86 
This should be all for now.

devemux86:
@Sublimis 

I'm thinking about whether we should make a new release, since so many improvements have been implemented.
Or do you have more work planned?

Sublimis:
More work! The labels are still not 100% perfect 😅 Maybe a day or two, if everything goes well. Also, the hill shading improvements have been completed for some time now, and are basically waiting to be pushed (this is a big commit, but tested a lot by now, and works great). After that yes, all of this will probably justify a new release.

devemux86:
@Sublimis 

No problem! It is better small pull requests step by step to understand the changes in history.

Then hill shading can come modifying only its own files so we have a clear implementation. :slightly_smiling_face:

devemux86:
@Sublimis 

No problem! It is better small pull requests step by step to understand the changes in history.

Then hill shading can come modifying only its own files, so we have a clear implementation. :slightly_smiling_face:

Sublimis:
Those h.s. changes couldn't be made in small steps, as there are changes to interfaces etc., you'll see 🙂 But yes, it mostly touches its own files, except for some small changes to GraphicContext and GraphicFactory just to add the color feature.

Sublimis:
> I may not fully understand the rationale behind this change,

IIRC, there were some abuses of the "always" property in default rendering themes that caused unwanted overlapping of displayed items. No one ever mentioned using the property to merge icons, so that was probably not the original intention.

> but if there are limitations with `Display.IFSPACE`, perhaps we could consider introducing a new parameter, something like `symbol/label-margin` - to define the minimum distance allowed between symbols or labels. This could help avoid overlaps and preserve clarity.

IMHO something like this should be the way to go if this functionality is to be implemented in Mapsforge, and `Display.ALWAYS` should be left as is.

devemux86:
@voldapet @Sublimis 

> No one ever mentioned using the property to merge icons, so that was probably not the original intention.

Indeed this seems like a new function.

Anyway, if you all read the discussion [above](https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-2517102211), could we allow `always` to work as before,
but enabling it via a flag in the `Parameters` class for those who need it?

_It is important that the existing Mapsforge / OpenAndroMaps / Freizeitkartemap themes
continue to work as they do now without the need for changes._

devemux86:
@voldapet @Sublimis 

> it was possible to "merge" a background and foreground symbol to create a unified symbol.

> No one ever mentioned using the property to merge icons, so that was probably not the original intention.

Indeed this seems like a new function.

Anyway, if you all read the discussion [above](https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-2517102211), could we allow `always` to work as before,
but enabling it via a flag in the `Parameters` class for those who need it?

_It is important that the existing Mapsforge / OpenAndroMaps / Freizeitkartemap themes
continue to work as they do now without the need for changes._

devemux86:
@voldapet @Sublimis 

If someone can provide a pull request with the `Display.FORCE` theme option or something similar,
I can review it.

Sublimis:
> If someone can provide a pull request with a new `Display.FORCE` theme option or something similar, I can review it.

Can do. Please allow a few days.

A question to @voldapet regarding multilayer icons: Is there some deeper reason why you are delegating this to Mapsforge instead of implementing it in your application? Even with the `FORCED` option, this still somehow feels like a hack (in the sense that it seems like it could break again in the future for other currently unknown and unforeseen reasons).

voldapet:
@Sublimis @devemux86 
I think we (@menion ) can also prepare the pull request for `Display.FORCE`. Just may take some time due to vacations and other commitments.

> Is there some deeper reason why you are delegating this to Mapsforge instead of implementing it in your application? 
I'm not sure I can understand. A few years ago, we maintained our own fork of Mapforge within the app. However, this approach created complications for the community - particularly around theme compatibility and other inconsistencies—so we ultimately chose to align with the official Mapsforge library

As for why we "merge" icons in Mapsforge, there are two main reasons:
- to simplify the theme. `osmc:symbol` can have thousands of combinations for foreground and background, which would otherwise require thousands of individual `<rule>` elements.
- to ensure that icons (merged) are displayed just near to another one (see the image in my previous [post](https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-3046372969) where icons "touch" each other
- we also considered using “shields” (icons with embedded text) to represent road numbers.

Sublimis:
Ok, thanks for the info. We've started working on this, PR will probably be ready by tomorrow at the latest.

PR #1588, opened 2024-12-09
Detailing the prevention of non-deterministic labels

Sublimis:
1. Improve `DirectRenderer`, i.e. make labels that are rendered directly on tiles (when using `MapDataStoreLabelStore`) deterministic with this one change: Omit drawing labels that span multiple tiles **and** clash with labels of similar priority.
2. Improve labels/symbols clash detection (using clash rectangles, not just absolute boundary).
3. Improve invisible labels/symbols handling (remove them before processing the rest).
4. `CanvasRasterer.renderContext` reference (it's natural, because `CanvasRasterer` only lives inside `RenderContext`).
5. Improve `RenderContext` fields encapsulation.
6. Tests for `LayerUtil.collisionAndContestingFreeOrdered()` and `LayerUtil.collisionFreeOrdered()`.

After implementing points 1. and 2., not a single label tear was observed when using the `DirectRenderer` method. It seems that the problem with non-deterministic labels has finally been completely solved. Yay!

Note that point 1. didn't change the way the current implementation of `DatabaseRenderer` handles tile-rendered labels when using `TileBasedLabelStore`.

We hope you don't mind points 4. and 5. It would be a pity to revert them because they make the code more future-proof and OOP-compliant (and less error-prone).

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
@@ -31,7 +31,7 @@ public abstract class PointTextContainer extends MapElementContainer {
     protected static final boolean DEBUG_CLASH_BOUNDS = false;
 
     public final double horizontalOffset;
-    public final boolean isVisible;
+    public final boolean isNotVisible;
     public final int maxTextWidth;
     public final Paint paintBack;
     public final Paint paintFront;
@@ -39,8 +39,6 @@ public abstract class PointTextContainer extends MapElementContainer {
     public final SymbolContainer symbolContainer;
     public final String text;
     public final double verticalOffset;
-    public volatile double clashRotationDegrees;
-    public volatile Rectangle clashRect;
 
     /**
      * Create a new point container, that holds the x-y coordinates of a point, a text variable, two paint objects, and
@@ -59,7 +57,7 @@ public abstract class PointTextContainer extends MapElementContainer {
         this.position = position;
         this.horizontalOffset = horizontalOffset;
         this.verticalOffset = verticalOffset;
-        this.isVisible = !this.paintFront.isTransparent() || (this.paintBack != null && !this.paintBack.isTransparent());
+        this.isNotVisible = super.isNotVisible() || (this.paintFront.isTransparent() && (this.paintBack == null || this.paintBack.isTransparent()));
     }
 
     @Override
@@ -68,18 +66,19 @@ public abstract class PointTextContainer extends MapElementContainer {
         if (Display.ALWAYS == this.display ^ Display.ALWAYS == other.display) {
             return false;
         }
+
+        Rectangle rect1 = this.getClashRect(rotation);
+        Rectangle rect2 = other.getClashRect(rotation);
+
+        if (rect1 != null && rect2 != null && rect1.intersects(rect2)) {
+            return true;
+        }
+
         if (!(other instanceof PointTextContainer)) {
             return false;
         }
 
         PointTextContainer ptc = (PointTextContainer) other;
-
-        Rectangle rect1 = getClashRect(this, rotation);
-        Rectangle rect2 = getClashRect(ptc, rotation);
-
-        if (rect1 != null && rect2 != null && rect1.intersects(rect2)) {
-            return true;
-        }
 
         if (this.text.equals(ptc.text) && this.xy.distance(ptc.xy) < 200) {
             return true;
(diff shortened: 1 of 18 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
```diff
@@ -67,7 +69,7 @@ public class DatabaseRenderer extends StandardRenderer {
      * @param hillsRenderConfig the hillshading setup to be used (can be null).
      */
     public DatabaseRenderer(MapDataStore mapDataStore, GraphicFactory graphicFactory, TileCache tileCache,
-                            TileBasedLabelStore labelStore, boolean renderLabels, boolean cacheLabels,
+                            LabelStore labelStore, boolean renderLabels, boolean cacheLabels,
```
devemux86, line 72: @Sublimis Why is this change?
As it is always a `TileBasedLabelStore`.

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
```diff
@@ -31,16 +31,14 @@ public abstract class PointTextContainer extends MapElementContainer {
     protected static final boolean DEBUG_CLASH_BOUNDS = false;
 
     public final double horizontalOffset;
-    public final boolean isVisible;
+    public final boolean isNotVisible;
```
devemux86, line 34: @Sublimis `isVisible` is more understandable and already used. Why change it?

mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DatabaseRenderer.java
Sublimis: It's not always -- even in the current master it can be `null`. `DatabaseRenderer` does not depend on this being a `TileBasedLabelStore`. Current implementation of `processLabels()` also does not *depend* on a particular `LabelStore` implementation, it only assumes there's a `LabelStore` that contains some labels, nothing more. It can even run with `labelStore == null`.

And now that `DirectRenderer` extends `DatabaseRenderer` in this PR, LabelStore implementation can also be `MapDataStoreLabelStore`, so it feels natural to generalize things. If this is really something you want to avoid we can for example create a new interface extending `LabelStore` that would define the `TileBasedLabelStore.storeMapItems()`. Otherwise too much code would get duplicated in `DirectRenderer` (there would be no point in extending `DatabaseRenderer` if it can only work with `TileBasedLabelStore`).

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/RenderContext.java
```diff
@@ -43,15 +46,15 @@ public class RenderContext {
     public final CanvasRasterer canvasRasterer;
 
     // Data generated for the rendering process
-    private List<List<ShapePaintContainer>> drawingLayers;
-    public final List<MapElementContainer> labels;
-    public final List<List<List<ShapePaintContainer>>> ways;
+    private List<List<ShapePaintContainer>> drawingLayer;
```
devemux86, line 49: @Sublimis Why in the singular?

mapsforge-core/src/main/java/org/mapsforge/core/mapelements/PointTextContainer.java
Sublimis: Two reasons: 1) `isVisible()` was only ever used in negation, cluttering the code, 2) `isVisible()` returning `true` doesn't really mean much, and can be misleading, because it does not mean the item is visible like the name implies. Like the javadoc of the superclass method `isNotVisible()` says: "@return {@code true} if this is definitely not visible; {@code false} if it may or may not be visible"

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/RenderContext.java
Sublimis: https://github.com/mapsforge/mapsforge/blob/dc4707532c3ee0cd9af8862ddca74ac9f32c8033/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/RenderContext.java#L68

In fact this is also a typo, it should also be singular

devemux86: > In fact this is also a typo, it should also be singular

@Sublimis Then we should also change this.

Sublimis:
By the end of the week we'll also push the latest hill shading improvements, and that should be all from us for the foreseeable future 🙂

devemux86:
@Sublimis Thanks!

PR #1592, opened 2024-12-16
Adaptive hill shading

Sublimis:
* Adaptive hill shading (`AdaptiveClasyHillShading`): Hill shading can now be rendered efficiently at all zoom levels (screenshots below).
    - Adaptive hill shading algorithm autonomously decides min/max zoom levels by default (others continue to work as before)
    - It will dynamically decide on the resolution and quality of the output depending on the display parameters, to maximize efficiency
    - Min/max zoom levels can be enforced via code
* Hill shading color feature. Color for the hill shading can now be set via render theme, or overriden by the code.
* New `<hillshading>` render theme element attribute: `"color"`.
* Deprecated the `"always"` attribute of the `<hillshading>` render theme element, it's obsolete.
* Significantly improved `HgtCache`: It will now constraint based on memory usage and also prevent unconstrained cache hammering and resulting excessive memory usage and possible OOM exceptions in extreme situations.
* `HgtFileInfo` class was made a standalone class.
* `AbsShadingAlgorithmDefaults` name was simplified (renamed) to `AShadingAlgorithm`.
* Hill shading bitmaps are now synchronized where needed (prevents various artifacts, at least on Android).
* Fix (AWT): Hill shading does not use `AwtLuminanceShadingComposite` any more, it's obsolete. It was used in the past to reduce the effect of the "neutral hills" gray overlay, which is not present any more. This change also improves performance of the hill shading on AWT, and restores symmetry with Android.
* Multithreaded indexing of DEM files

395217943-a4fbd776-3fef-42ef-8873-a305b49c0b19


mapsforge-core/src/main/java/org/mapsforge/core/graphics/ABitmap.java
```diff
@@ -0,0 +1,25 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.mapsforge.core.graphics;
+
+public abstract class ABitmap implements Bitmap {
```
devemux86, line 17: @Sublimis Let's use a better name, like `BaseBitmap`.

mapsforge-core/src/main/java/org/mapsforge/core/graphics/ABitmap.java
```diff
@@ -0,0 +1,25 @@
...
+
+   protected final Object Mutex = new Object();
```
devemux86, line 19: @Sublimis Name should start in lower case.

mapsforge-core/src/main/java/org/mapsforge/core/model/Tile.java
```diff
@@ -107,8 +108,9 @@ public static int getMaxTileNumber(byte zoomLevel) {
      */
     public final byte zoomLevel;
 
-    private BoundingBox boundingBox;
-    private Point origin;
+    private volatile BoundingBox boundingBox;
+    private volatile Point origin;
+    private transient final Object mSync = new Object();
```
devemux86, line 113: @Sublimis We do not use the prefix `m` in names in Mapsforge.

mapsforge-core/src/main/java/org/mapsforge/core/util/MercatorProjection.java
```diff
@@ -216,9 +210,10 @@ public static double latitudeToPixelY(double latitude, byte zoomLevel, int tileS
      * @return the pixel Y coordinate of the latitude value.
      */
     public static double latitudeToPixelY(double latitude, long mapSize) {
-        double sinLatitude = Math.sin(latitude * (Math.PI / 180));
+        double sinLatitude = Math.sin(Math.toRadians(latitude));
         // FIXME improve this formula so that it works correctly without the clipping
         double pixelY = (0.5 - Math.log((1 + sinLatitude) / (1 - sinLatitude)) / (4 * Math.PI)) * mapSize;
+//        double pixelY = (0.5 - Math.log(Math.tan(Math.toRadians(45 + latitude / 2))) / (2 * Math.PI)) * mapSize;
```
devemux86, line 216: @Sublimis What is this commented code?

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -49,8 +50,12 @@ public enum SymbolScaling {ALL, POI}
 
     /**
      * Use AwtLuminanceShadingComposite or AlphaComposite.
+     * (2024: Deprecated, should not be used)
      */
-    public static boolean LUMINANCE_COMPOSITE = true;
+    public static final boolean LUMINANCE_COMPOSITE = false;
```
devemux86, line 55: @Sublimis Parameters cannot be final. Either keep it non final or remove the `LUMINANCE_COMPOSITE`.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -49,8 +50,12 @@ public enum SymbolScaling {ALL, POI}
...
+
+    public static final int HILLSHADING_MAGNITUDE_DEFAULT = 128;
+    public static final int HILLSHADING_COLOR_DEFAULT = 0xff000000;
```
devemux86, line 58: @Sublimis These are not parameters, but constants. They should be moved to the class that uses them.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -64,6 +69,8 @@ public enum SymbolScaling {ALL, POI}
      */
     public static int NUMBER_OF_THREADS = Runtime.getRuntime().availableProcessors() + 1;
 
+    public static final long MAX_MEMORY_MB = Runtime.getRuntime().maxMemory() / 1000 / 1000;
```
devemux86, line 72: @Sublimis This is not a parameter, but a calculated constant. It should be moved to the class that uses it.

devemux86, line 72: @Sublimis This is not a parameter, but a calculated constant. It should be moved to the class that uses it.

Also, shouldn't the calculation be done with `1024'?

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/MemoryCachingHgtReaderTileSource.java
```diff
@@ -17,33 +17,52 @@
...
+    // To prevent cache starvation
+    protected final int CacheMinCount = 1;
+
+    // Each HGT file contains 1° x 1° DEM data, so there can be at most 180 x 360 HGT files.
+    // The actual number of files is smaller because there is currently no bathymetric data.
+    protected final int CacheMaxCount = 360 * 180;
+
+    // One 1" HGT file converted to a same-sized bitmap is about 13 MB, for high-quality this is 52 MB.
+    // For ultra-low-quality while rendering wide zoom in adaptive mode, bitmap size per 1" HGT file can be as low as a few hundred bytes.
+    protected final long CacheMaxBytes = Parameters.MAX_MEMORY_MB * 1000 * 1000 / 10;
```
devemux86, line 42: @Sublimis Shouldn't the calculation be done with `1024'?

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/RenderThemeHandler.java
```diff
@@ -282,8 +284,8 @@ else if ("overlay".equals(qName)) {
                 byte minZoom = 9;
                 byte maxZoom = 17;
                 byte layer = 5;
-                short magnitude = 128;
-                boolean always = false;
+                int magnitude = Parameters.HILLSHADING_MAGNITUDE_DEFAULT;
```
devemux86, line 287: @Sublimis `RenderThemeHandler.HILLSHADING_MAGNITUDE` should also become `int`.

Sublimis: It is an int, `public static final int HILLSHADING_MAGNITUDE_DEFAULT = 128;` ?

mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/HillshadingMapViewer.java
```diff
@@ -96,7 +99,6 @@ protected void createLayers() {
     }
 
     private void customizeConfig(MemoryCachingHgtReaderTileSource hillTileSource) {
-        hillTileSource.setEnableInterpolationOverlap(true);
```
devemux86, line 99: @Sublimis This empty method should be removed.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
Sublimis: What if more than one class wants to use a constant, shouldn't there be some class holding constants?
Can we create it?
https://en.wikipedia.org/wiki/Megabyte

mapsforge-core/src/main/java/org/mapsforge/core/util/MercatorProjection.java
Sublimis: Some leftover, sorry, while trying to fix something. It is just another way of doing the exact same calculation in the line above.

mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/rule/RenderThemeHandler.java
devemux86: @Sublimis I wrote about the existing in `RenderThemeHandler` to become an `int`:
```java
public static short HILLSHADING_MAGNITUDE = -1;
```

resources/renderTheme.xsd
```diff
@@ -321,16 +321,23 @@
...
+                    "magnitude" attribute - you can use both to control the opacity of the hill
+                    shading. They are multiplied, so for example a magnitude of 128 and a color of
+                    #ff000000 would result in an opacity of 50%, while a magnitude of 128 and a
+                    color of #80000000 would result in an opacity of 25% (0.5 * 0.5 = 0.25).
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
         <!-- customization unlikely to make any sense -->
         <xs:attribute name="layer" default="5" type="xs:unsignedByte" use="optional" />
+        <xs:attribute name="always" default="false" type="xs:boolean" use="optional">
```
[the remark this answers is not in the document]
Sublimis: This would immediately break the validity of all rendering themes that already use the "always" attribute, wouldn't it? That's why there should be some grace period...

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
devemux86: @Sublimis `Parameters` is not the class to hold constants.
We can create a `Constants` class for such purpose.

resources/renderTheme.xsd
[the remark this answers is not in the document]
devemux86: @Sublimis It doesn't matter, as render themes still work.
XML validation can be checked by the theme authors.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -111,6 +107,19 @@ public enum SymbolScaling {ALL, POI}
         POLYGON_EXCEPTIONS.add(new Tag("freizeitkarte", "land"));
     }
 
+    public static final class Constants {
```
devemux86, line 110: @Sublimis Let's make `Constants` a separate class.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -111,6 +107,19 @@ public enum SymbolScaling {ALL, POI}
...
+
+        public static final int HILLSHADING_MAGNITUDE_DEFAULT = 128;
+        public static final int HILLSHADING_COLOR_DEFAULT = 0xff000000;
+        public static final long MAX_MEMORY_MB = Runtime.getRuntime().maxMemory() / 1000 / 1000;
+
+        /**
+         * Use AwtLuminanceShadingComposite or AlphaComposite.
+         * (2024: Deprecated, should not be used)
+         */
+        public static final boolean LUMINANCE_COMPOSITE = false;
```
devemux86, line 120: @Sublimis As it cannot be used externally, we can remove it.
And also the `AwtLuminanceShadingComposite` class.

mapsforge-core/src/main/java/org/mapsforge/core/util/Parameters.java
```diff
@@ -111,6 +107,19 @@ public enum SymbolScaling {ALL, POI}
         POLYGON_EXCEPTIONS.add(new Tag("freizeitkarte", "land"));
     }
 
+    public static final class Constants {
+
+        public static final int HILLSHADING_MAGNITUDE_DEFAULT = 128;
+        public static final int HILLSHADING_COLOR_DEFAULT = 0xff000000;
```
devemux86, line 113: @Sublimis Also the hillshading constants belong in the 1 class that uses them.
They are not used anywhere else (excluding possible tests).

devemux86:
@Sublimis Thanks, very nice work!

Sublimis:
Should we do another PR for those final details about constants etc. or do you think it's fine the way it is now?
Thanks!

devemux86:
@Sublimis I separated the `Constants` class in #1593.

Sublimis:
Screenshots update

devemux86:
> Screenshots update

@Sublimis Thanks, very nice! :+1:

devemux86:
@Sublimis 

There are reports for ANR or crash with the hillshading lately.

Although we have not seen anything strange in [Cruiser](https://github.com/devemux86/cruiser) so far.

- https://github.com/mapsforge/mapsforge/issues/1595

- https://github.com/cgeo/cgeo/issues/16429

ztNFny:
> * [Unified Mapsforge] ANR toggling hillshading cgeo/cgeo#16429

fyi, at least that one is from before Adaptive hillshading was merged

@Sublimis however I found some other issues with adaptive hillshading, documented in https://github.com/cgeo/cgeo/issues/16438#issuecomment-2558421987, maybe you could take a look / make a statement on that?

PR #1597, opened 2024-12-24
Hill shading improvements

Sublimis:
How about something like this?

* Interrupt and destroy hill shading thread pools from the `MapWorkerPool` (maybe #1595)
* Fix: `AdaptiveClasyHillShading.setAdaptiveZoomEnabled(false)` not entirely respected (https://github.com/cgeo/cgeo/issues/16438#issuecomment-2558517632)
* Threaded hill shading timing debug feature. Let `AThreadedHillShading.isDebugTiming()` return `true` to see the results:

```java
/**
 * @return {@code true} to measure and output rendering times per file.
 * @see #isDebugTimingSequential()
 */
protected boolean isDebugTiming() {
    return false;
}

/**
 * @return {@code true} to process one file at a time for more accurate timings. Note: Rendering will be slower.
 * @see #isDebugTiming()
 */
protected boolean isDebugTimingSequential() {
    return true;
}
```

The results will be presented like this:

```java
final double delayMs = Math.round(delayNano / 1e5) / 10.;

final String debugTag = this.getClass().getSimpleName() + "-R" + mReadingThreadsCount + "-C" + mComputingThreadsCount + "-E" + ElementsPerComputingTask + "-HQ" + (isHighQuality ? 1 : 0) + "-Z" + (zoomLevel < 10 ? "0" : "") + zoomLevel + " T: " + delayMs + " ms";

System.out.println(debugTag);
```

image

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
@@ -553,7 +553,7 @@ public abstract class AThreadedHillShading extends AShadingAlgorithm {
 
     @Override
     protected byte[] convert(InputStream inputStream, int dummyAxisLen, int dummyRowLen, int padding, int zoomLevel, double pxPerLat, double pxPerLon, HgtFileInfo hgtFileInfo) throws IOException {
-        return doTheWork_(hgtFileInfo, false, padding, zoomLevel, pxPerLat, pxPerLon);
+        return convert(hgtFileInfo, false, padding, zoomLevel, pxPerLat, pxPerLon);
     }
 
     /**
@@ -565,21 +565,15 @@ public abstract class AThreadedHillShading extends AShadingAlgorithm {
      * @param pxPerLon      Tile pixels per degree of longitude (to determine shading quality requirements)
      * @return
      */
-    protected byte[] doTheWork_(final HgtFileInfo hgtFileInfo, boolean isHighQuality, int padding, int zoomLevel, double pxPerLat, double pxPerLon) {
+    protected byte[] convert(final HgtFileInfo hgtFileInfo, boolean isHighQuality, int padding, int zoomLevel, double pxPerLat, double pxPerLon) {
         final byte[] output;
 
-        if (false == isDebugTiming()) {
+        if (!isDebugTiming()) {
             output = doTheWork(hgtFileInfo, isHighQuality, padding, zoomLevel, pxPerLat, pxPerLon);
         } else {
             final long startTs, finishTs;
 
-            if (false == isDebugTimingSequential()) {
-                startTs = System.nanoTime();
-
-                output = doTheWork(hgtFileInfo, isHighQuality, padding, zoomLevel, pxPerLat, pxPerLon);
-
-                finishTs = System.nanoTime();
-            } else {
+            if (isDebugTimingSequential()) {
                 // We want to process one file at a time for more accurate timings
                 synchronized (mDebugSync) {
                     startTs = System.nanoTime();
@@ -588,6 +582,12 @@ public abstract class AThreadedHillShading extends AShadingAlgorithm {
 
                     finishTs = System.nanoTime();
                 }
+            } else {
+                startTs = System.nanoTime();
+
+                output = doTheWork(hgtFileInfo, isHighQuality, padding, zoomLevel, pxPerLat, pxPerLon);
+
+                finishTs = System.nanoTime();
             }
 
             final long delayNano = finishTs - startTs;
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
@@ -170,13 +170,13 @@ public abstract class AThreadedHillShading extends AShadingAlgorithm {
     protected final int mActiveTasksCountMax;
 
     /**
-     * Static thread pools shared by all tasks.
+     * Thread pools.
      */
     protected final AtomicReference<HillShadingThreadPool> mReadThreadPool = new AtomicReference<>(null);
     protected final AtomicReference<HillShadingThreadPool> mCompThreadPool = new AtomicReference<>(null);
 
     /**
-     * Stop signal flag, indicating that processing should be stopped as soon as possible (not used by default).
+     * Stop signal flag, indicating that processing should be stopped as soon as possible.
      *
      * @see #isNotStopped()
      */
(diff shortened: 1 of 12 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
```diff
@@ -553,7 +553,52 @@ protected int getStrideFactor(int inputAxisLen, int outputAxisLen) {
...
+    /**
+     * @param hgtFileInfo   HGT file info
+     * @param isHighQuality When {@code true}, a unit element is 4x4 data points in size instead of 2x2, for better interpolation possibilities.
+     * @param padding       Padding of the output, useful to minimize border interpolation artifacts (no need to be larger than 1)
+     * @param zoomLevel     Zoom level (to determine shading quality requirements)
+     * @param pxPerLat      Tile pixels per degree of latitude (to determine shading quality requirements)
+     * @param pxPerLon      Tile pixels per degree of longitude (to determine shading quality requirements)
+     * @return
+     */
+    protected byte[] doTheWork_(final HgtFileInfo hgtFileInfo, boolean isHighQuality, int padding, int zoomLevel, double pxPerLat, double pxPerLon) {
```
devemux86, line 568: @Sublimis `doTheWork` and `doTheWork_` are confusing.
Please avoid the `_` and name the methods differently.
`doTheWork` can become private as is not called elsewhere.

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
```diff
@@ -553,7 +553,52 @@ protected int getStrideFactor(int inputAxisLen, int outputAxisLen) {
...
+        final byte[] output;
+
+        if (false == isDebugTiming()) {
```
devemux86, line 571: @Sublimis Use `!isDebugTiming()`.

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
```diff
@@ -553,7 +553,52 @@ protected int getStrideFactor(int inputAxisLen, int outputAxisLen) {
...
+            output = doTheWork(hgtFileInfo, isHighQuality, padding, zoomLevel, pxPerLat, pxPerLon);
+        } else {
+            final long startTs, finishTs;
+
+            if (false == isDebugTimingSequential()) {
```
devemux86, line 576: @Sublimis Use `!isDebugTimingSequential()`.

devemux86:
@mg4gh

I merged this pull request in branch https://github.com/mapsforge/mapsforge/tree/issue_1597.
Can you test it?

Test it also with `AndroidGraphicFactory.clearResourceMemoryCache()`.

Sublimis:
@devemux86 Thanks for the comments. But what about the important stuff, can anyone confirm if this helps fix the #1595  from the first point? @mg4gh ?

Sublimis:
> @devemux86 Thanks for the comments. But what about the important stuff, can anyone confirm if this helps fix the #1595 from the first point? @mg4gh ?

(Because we cannot recreate that issue.)
However, this does successfully quit hill shading threads, that part has been tested and confirmed.

mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/AThreadedHillShading.java
Sublimis: No privacy in a non-final class, sorry

Sublimis:
@devemux86 A somewhat related question. Is the code that makes `MemoryCachingHgtReaderTileSource` mutable useful for anything and still needed?

https://github.com/mapsforge/mapsforge/blob/dba0ce920bb25c0ce3d3a14da9c83e8fd37ebf24/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/MemoryCachingHgtReaderTileSource.java#L119-L122

https://github.com/mapsforge/mapsforge/blob/dba0ce920bb25c0ce3d3a14da9c83e8fd37ebf24/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/MemoryCachingHgtReaderTileSource.java#L137-L139

https://github.com/mapsforge/mapsforge/blob/dba0ce920bb25c0ce3d3a14da9c83e8fd37ebf24/mapsforge-map/src/main/java/org/mapsforge/map/layer/hills/MemoryCachingHgtReaderTileSource.java#L92-L96

devemux86:
@Sublimis Probably not.

mg4gh:
Sorry for not being more responsive, but I will not be able to do much before Friday afternoon. Then I will do the test as suggested by @devemux86

mg4gh:
Test with mapsforge/issue_1597 software, with `AndroidGraphicFactory.clearResourceMemoryCache()` and 1 render thread I got following crash: [20241225_222803_crash.log](https://github.com/user-attachments/files/18247267/20241225_222803_crash.log)
Further environment: device Pixel 6 with germany map (from openandromaps), zoom level 9, area south of heidelberg.
Procedure: Switch Setting for Hillshading on/off, which triggers a recreate of the app - after a few times it crashed.
The pool-25-thread-1 (which is mentioned in the last line) was created before the last recreate.

Sublimis:
There are a lot of things happening here, RenderThemeHandler is parsing render theme, hill shading is reading from ZIPs and rendering in high quality, your app is doing something, probably some map animation is going on, but the exact line where the error occurs is still not clear. As you said, this probably has nothing to do with hill shading, other than the fact that it's easier to recreate the error when hill shading is actively using resources. Can you somehow pinpoint the exact line of code where this is happening, so we know which bitmap the error is referring to? Since the error is coming from HWUI, it's probably a map View issue, but... (One thing to note, render theme parsing cannot be interrupted, it is not implemented. Maybe this could help.)

devemux86:
> Switch Setting for Hillshading on/off, which triggers a recreate of the app

This is important information and is actually a bad practice.
Changing layers on the map should not destroy / recreate the app!
Just remove / add the layers.

devemux86:
Or if  want to constantly recreate the app, do not use `AndroidGraphicFactory.clearResourceMemoryCache()`
and let the system handle the unused bitmaps.

devemux86:
Or if  want to constantly recreate the app, do not use `AndroidGraphicFactory.clearResourceMemoryCache()`
and let the system handle the unused bitmaps, as it happens on the desktop.

mg4gh:
> @mg4gh
> 
> I merged this pull request in branch https://github.com/mapsforge/mapsforge/tree/issue_1597. Can you test it?
> 
> Test it also with `AndroidGraphicFactory.clearResourceMemoryCache()`.

The test was done according to this request. And I tried to check whether I can reproduce the failure. As the `AndroidGraphicFactory.clearResourceMemoryCache()` is part of the destroy, I triggered the destroy by an action that results in a recreate. I understand that I should change my app behaviour and work without recreate - but this will take some time and it is not so easy for me.

Anyway I tested the same scenario without hillshading, but with multiple render threads, and  the result is - as expected - still the same. Without recreation of the activity this seems to be no problem. Still for me it feels not the right way, that resources after the stop of the MapWorkerPool continue their work as render threads.

devemux86:
@Sublimis Is this pull request still useful?
Should we merge it or more work is needed?

Sublimis:
This is good in itself. Btw just wanted to tell you that we're preparing some improvements to adaptive hill shading, mostly performance, but also additional customizations. Can we send it here, or should it be another PR? Tomorrow at the latest, if additional testing will be needed.

devemux86:
@Sublimis Let's merge this and you can create a new pull request for additional improvements.

devemux86:
@Sublimis Thanks!

devemux86:
@Sublimis 

I tried setting a hillshading color in Motorider dark theme. It does not seem to work with exception:
```java
org.mapsforge.map.layer.renderer.DatabaseRenderer executeJob
WARNING: java.util.concurrent.ExecutionException: java.lang.NumberFormatException: For input string: "#ffff0000"
```

Sublimis:
Thanks, it should be `getColor()`, not `parseNonNegativeInteger()`, of course, fixed it now for the next PR.

Sublimis:
Do you want a little PR just to get this fixed right away?

devemux86:
> Do you want a little PR just to get this fixed right away?

@Sublimis Sure, we better do this.

Sublimis:
Just make sure to comment out lines like `hillsConfig.setColor(0xff000000);` if using the samples if you don't want the theme color to be overridden.

devemux86:
@Sublimis And I wondered why it did not work!

We probably do not need these lines in samples, as they set default values and can cause such confusion:
```
hillsConfig.setMagnitudeScaleFactor(1);
hillsConfig.setColor(0xff000000);
```

Sublimis:
It would probably be best to keep them so people know they exist, but commented out, or keep them as they are but put a comment above saying this overrides the theme values...

devemux86:
> but commented out

@Sublimis Let's comment them, so we can make tests without having to also change the code.

PR #1601, opened 2024-12-27
Fix: NumberFormatException when setting hill shading color

Sublimis:
* Fix: NumberFormatException when setting hill shading color in render theme


devemux86:
@Sublimis Thanks!

PR #1603, opened 2024-12-27
Add ZOrderGroupLayer to Mapsforge

eddiemuc:
Adds class ZOrderGroupLayer to Mapsforge, including a usage example for android in OverlayMapViewer:

image


devemux86:
@eddiemuc Thanks for this interesting contribution and the sample!

PR #1611, opened 2025-01-04
Fix: AwtGraphicFactory.createMonoBitmap() hill shading color sync

Sublimis:
#1610


Sublimis:
This should probably be rock solid now.

devemux86:
@Sublimis Thanks for looking into this!

PR #1612, opened 2025-01-05
Adaptive hill shading: Remove the default maximum zoom limits

Sublimis:
* Remove the default maximum zoom limits from the Adaptive hill shading algorithm

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
@@ -4,7 +4,6 @@
  * Copyright 2015-2020 devemux86
  * Copyright 2015 Andreas Schildbach
  * Copyright 2016 mikes222
- * Copyright 2025 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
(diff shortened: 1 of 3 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/model/MapViewPosition.java
```diff
@@ -28,6 +29,9 @@
 
 public class MapViewPosition extends Observable implements Persistable {
 
+    public static final int ZOOM_MIN_DEFAULT = 0;
+    public static final int ZOOM_MAX_DEFAULT = 22;
```
devemux86, line 33: @Sublimis Please do not put such custom zoom limits.
There are apps that use the map on large zoom levels.

Users can limit the max zoom with the related method.

Sublimis: Can you leave this commit out, or should we push another one that will revert it?
Is the unlimited maximum zoom in the app samples also intentional or can we limit it there? To prevent zooming indefinitely, while the scale remains at "1 meter". But it doesn't really matter, it's just an inconvenience.

devemux86: @Sublimis Better make a new pull request if cannot revert the commit?

devemux86:
@Sublimis Thanks!

PR #1629, opened 2025-01-17
Ways deduplication performance improvement

Sublimis:
* Deduplication performance was improved just for ways (pois not used).
* Tested using three `Germany_oam.osm.map` files (just renamed, so exactly 3x duplication).
* `MapWorkerPool.DEBUG_TIMING = true`, starting with München area on scale = 20km.
* Tested on AWT only.

`DEDUPLICATE` (improved) took **2.3** seconds on average (±0.1).
`RETURN_ALL` took **3.1** seconds on average (±0.1).

The performance of the legacy DEDUPLICATE method was tested later and found to be comparable to RETURN_ALL, thus significantly slower (unsurprisingly).

Android wasn't tested, but the relative results could be even better there, given that mobile devices have weaker GPUs. Not sure, though, maybe someone could do some testing on Android.

This makes the `DEDUPLICATE` method a good candidate for the recommended and default behavior, possibly mitigating the need for map bounding polygons (#1608).

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapReadResult.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapReadResult.java
@@ -2,6 +2,7 @@
  * Copyright 2010, 2011, 2012, 2013 mapsforge.org
  * Copyright 2014-2015 Ludwig M Brinckmann
  * Copyright 2015-2022 devemux86
+ * Copyright 2025 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -17,7 +18,9 @@
 package org.mapsforge.map.datastore;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
@@ -30,7 +33,6 @@ public class MapReadResult {
      * Hash codes.
      */
     private final Set<Integer> hashPois = new HashSet<>();
-    private final Set<Integer> hashWays = new HashSet<>();
 
     /**
      * True if the read area is completely covered by water, false otherwise.
@@ -44,12 +46,14 @@ public class MapReadResult {
 
     /**
      * The read ways.
+     * LinkedHashSet is used to: 1) maintain element order, 2) maximize element removal performance when deduplicating.
      */
-    public List<Way> ways;
+    public final Set<Way> ways;
 
     public MapReadResult() {
         this.pointOfInterests = new ArrayList<>();
-        this.ways = new ArrayList<>();
+        // LinkedHashSet is used to: 1) maintain element order, 2) maximize element removal performance when deduplicating.
+        this.ways = new LinkedHashSet<>();
     }
 
     public void add(PoiWayBundle poiWayBundle) {
@@ -71,14 +75,35 @@ public class MapReadResult {
                     this.pointOfInterests.add(poi);
                 }
             }
-            for (Way way : other.ways) {
-                if (this.hashWays.add(way.hashCode())) {
-                    this.ways.add(way);
-                }
-            }
         } else {
             this.pointOfInterests.addAll(other.pointOfInterests);
-            this.ways.addAll(other.ways);
         }
+
+        this.ways.addAll(other.ways);
+    }
+
+    public MapReadResult deduplicate()
+    {
+        if (!this.ways.isEmpty()) {
+
+            final ArrayList<Way> list = new ArrayList<>(this.ways);
+
+            Collections.sort(list);
+
+            Way current = list.get(0);
+
+            for (int i = 1; i < list.size(); i++) {
+                Way way = list.get(i);
+                if (current.compareTo(way) == 0) {
+                    // Removing instead of building a new list because the expected number of duplicates is usually (much) less than 50%.
+                    this.ways.remove(way);
+                    continue;
+                }
+
+                current = way;
+            }
+        }
+
+        return this;
     }
 }
--- a/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
@@ -1,7 +1,7 @@
 /*
  * Copyright 2014-2015 Ludwig M Brinckmann
  * Copyright 2015-2022 devemux86
- * Copyright 2024 Sublimis
+ * Copyright 2024-2025 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -154,7 +154,7 @@ public class MultiMapDataStore extends MapDataStore {
             case RETURN_ALL:
                 return readLabels(tile, false);
             case DEDUPLICATE:
-                return readLabels(tile, true);
+                return readLabels(tile, true).deduplicate();
         }
         throw new IllegalStateException("Invalid data policy for multi map database");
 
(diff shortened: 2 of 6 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
```diff
@@ -154,7 +154,7 @@ public MapReadResult readLabels(Tile tile) {
             case RETURN_ALL:
                 return readLabels(tile, false);
             case DEDUPLICATE:
-                return readLabels(tile, true);
+                return readLabels(tile, true).deduplicate();
```
devemux86, line 157: @Sublimis Why calling extra method?
We already pass `deduplicate` true, so it should happen in `readLabels` like the pois.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapReadResult.java
```diff
@@ -71,14 +75,35 @@ public void add(MapReadResult other, boolean deduplicate) {
                     this.pointOfInterests.add(poi);
                 }
             }
-            for (Way way : other.ways) {
-                if (this.hashWays.add(way.hashCode())) {
-                    this.ways.add(way);
-                }
-            }
```
devemux86, line 78: @Sublimis New deduplicate should happen here like the pois.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapReadResult.java
```diff
@@ -71,14 +75,35 @@ public void add(MapReadResult other, boolean deduplicate) {
...
         } else {
             this.pointOfInterests.addAll(other.pointOfInterests);
-            this.ways.addAll(other.ways);
```
devemux86, line 81: @Sublimis Why? Without deduplicate, the ways will be empty.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/Way.java
```diff
@@ -53,50 +53,73 @@ public Way(byte layer, List<Tag> tags, LatLong[][] latLongs, LatLong labelPositi
...
-            return true;
-        } else if (!(obj instanceof Way)) {
-            return false;
+    public int hashCode() {
+        return System.identityHashCode(this);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return this == o;
```
devemux86, line 62: @Sublimis Is it enough check for `equals`?

mapsforge-map/src/main/java/org/mapsforge/map/datastore/Way.java
```diff
@@ -53,50 +53,73 @@ public Way(byte layer, List<Tag> tags, LatLong[][] latLongs, LatLong labelPositi
     }
 
     @Override
-    public boolean equals(Object obj) {
-        if (this == obj) {
-            return true;
-        } else if (!(obj instanceof Way)) {
-            return false;
+    public int hashCode() {
+        return System.identityHashCode(this);
```
devemux86, line 57: @Sublimis Is this hash code for speed?

mapsforge-map/src/main/java/org/mapsforge/map/datastore/Way.java
Sublimis: What do you mean?

devemux86: @Sublimis The previous `equals` made many checks.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
Sublimis: An additional method is required when you look at the algorithm. It doesn't really matter where exactly it gets called, as long as it's after all the reads and before returning the data to the caller. Now, using an extra parameter just for this is not productive. This parameter should go the way of the dodo once the algorithm is applied to POIs as well.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/Way.java
Sublimis: Yes, and this has been delegated to the compareTo() now. Please recall the flowchart: https://github.com/mapsforge/mapsforge/discussions/1608#discussioncomment-11749672

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapReadResult.java
devemux86: @Sublimis Please see [above](https://github.com/mapsforge/mapsforge/pull/1629#discussion_r1920376921).

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
devemux86: @Sublimis Then we should make the new method to work also on the pois and remove the parameter?

Now the method appears half implemented as the `deduplicate` parameter works only on the pois
and we need to call an extra method to do the previous work on the ways too.

Sublimis: Please let's do it in two steps. First to make this available to others for testing, and if all is good and accepted we'll implement pois as well. The functionality won't change anyway, and a new version is also not on the horizon :)

devemux86: @Sublimis Sure, users can test the PR in their builds.

But in order to merge it, we need the API to be complete.
This is public API that now a method does half work.

Sublimis: You mean the new public deduplicate method? Well it can be (package)protected as well, if that's your concern. Perhaps you could commit this to a new branch first, as we suggested earlier, if you don't want to keep this PR waiting? Until feedback is received, etc.

Sublimis: But why even bother? Anyone using a snapshot builds should be aware that they are using the "beta" version that could change in the next release, right?

devemux86: @Sublimis I mean the `MapReadResult.add(MapReadResult other, boolean deduplicate)`
that in the PR now only deduplicates the pois and needs and extra method for the ways.

Many users use the snapshots in their apps, so we cannot merge it in this state.
Unless the public `add` either works as before or we remove it and use the extra method.

devemux86: @Sublimis Why not using the old API and deduplicate (with the new algorithm) the ways inside the `add`
like before?

Sublimis: 👍 Let's then wait a few hours/days until we get some feedback from folks at #1608, @mg4gh ?

Sublimis: > @Sublimis Why not using the old API and deduplicate (with the new algorithm) the ways inside the `add` like before?

It feels more natural, less possibility of mistake, and it's usually better to be explicit about intentions in the code. Not a big deal though. The private `MultiMapDataStore.read...()` methods can keep their parameter (but why only `boolean` then, why not the `dataPolicy`?), however `MapReadResult.add(MapReadResult, boolean)` will definitely become obsolete.

devemux86:
I merged it in branch [issue_1629](https://github.com/mapsforge/mapsforge/tree/issue_1629) for testing.

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
Sublimis: > @Sublimis Why not using the old API and deduplicate (with the new algorithm) the ways inside the `add` like before?

Oh, you are talking about the MapReadResult.add()?? This method is called from the loop. While it could be done this way, it wouldn't be optimal, i.e. the resulting algorithm would be slower. It's much better to call the sort() once on a large collection, that many times on smaller collections. Mathematics.

PR #1633, opened 2025-01-19
MapDataStoreLabelStore.getVisibleItems read all map data, fix #1631

### LINKED ISSUE #1631: Label layer with MapDataStoreLabelStore shows less symbols
When we use a label layer with label cache (e.g. `LabelLayerUsingLabelCacheMapViewer`)
or we do not use a label layer (e.g. `SimplestMapViewer`), we see the **correct** first 2 images.

When we use a label layer with `MapDataStoreLabelStore` (e.g. `LabelLayerUsingMapDataStoreMapViewer`),
we see the **wrong** third image with much less symbols.

Was always the case or it is a result of the map label / symbol improvements in https://github.com/mapsforge/mapsforge/pull/1579 https://github.com/mapsforge/mapsforge/pull/1586 https://github.com/mapsforge/mapsforge/pull/1588 ?

cc @Sublimis

devemux86:
Fix #1631


Sublimis:
Unfortunately this has a significant performance penalty, almost to the point of being unusable, please see a workaround until a possible better solution is found: #1636

devemux86:
I have not seen any performance problems, but we can revert it.
This label layer with `MapDataStoreLabelStore` is not used and not recommended anyway because it has slow rendering.

devemux86:
See #1637

PR #1635, opened 2025-01-19
Ways and POIs deduplication performance and correctness improvement

Sublimis:
Deduplication performance and correctness improvement for ways **and POIs**.

* Tested using three `Germany_oam.osm.map` files (just renamed, so exactly 3x duplication), without POIs.
* `MapWorkerPool.DEBUG_TIMING = true`, starting with München area on scale = 20km.
* Tested on AWT only.

Results:

* `DEDUPLICATE` (improved) took **2.3** seconds on average (±0.1) to draw the map.
* `RETURN_ALL` took **3.1** seconds on average (±0.1) to draw the map.

The performance of the legacy DEDUPLICATE method was tested later and found to be comparable to RETURN_ALL, thus significantly slower (unsurprisingly).

An additional advantage of the new deduplication approach is correctness: This algorithm does not suffer from [hash collisions](https://en.wikipedia.org/wiki/Hash_collision), unlike the previous one. This becomes increasingly important as maps get bigger.

Android wasn't tested, but the relative results could be even better there, given that mobile devices have weaker GPUs. Not sure, though, maybe someone could do some testing on Android.

This makes the `DEDUPLICATE` method a good candidate for the recommended and default behavior, possibly mitigating the need for map bounding polygons (#1608).

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
@@ -1,7 +1,7 @@
 /*
  * Copyright 2014-2015 Ludwig M Brinckmann
  * Copyright 2015-2022 devemux86
- * Copyright 2024 Sublimis
+ * Copyright 2024-2025 Sublimis
  *
  * This program is free software: you can redistribute it and/or modify it under the
  * terms of the GNU Lesser General Public License as published by the Free Software
@@ -152,15 +152,15 @@ public class MultiMapDataStore extends MapDataStore {
                 }
                 return null;
             case RETURN_ALL:
-                return readLabels(tile, false);
+                return readLabelsPriv(tile);
             case DEDUPLICATE:
-                return readLabels(tile, true);
+                return readLabelsPriv(tile).deduplicate();
         }
         throw new IllegalStateException("Invalid data policy for multi map database");
 
     }
 
-    private MapReadResult readLabels(Tile tile, boolean deduplicate) {
+    private MapReadResult readLabelsPriv(Tile tile) {
         MapReadResult mapReadResult = new MapReadResult();
         boolean isTileFilled = false;
         for (MapDataStore mdb : mapDatabases) {
@@ -175,7 +175,7 @@ public class MultiMapDataStore extends MapDataStore {
                 }
                 boolean isWater = mapReadResult.isWater & result.isWater;
                 mapReadResult.isWater = isWater;
-                mapReadResult.add(result, deduplicate);
+                mapReadResult.add(result);
             }
 
             if (mdb.supportsFullTile(tile)) {
@@ -199,15 +199,15 @@ public class MultiMapDataStore extends MapDataStore {
                 }
                 return null;
             case RETURN_ALL:
-                return readLabels(upperLeft, lowerRight, false);
+                return readLabelsPriv(upperLeft, lowerRight);
             case DEDUPLICATE:
-                return readLabels(upperLeft, lowerRight, true);
+                return readLabelsPriv(upperLeft, lowerRight).deduplicate();
         }
         throw new IllegalStateException("Invalid data policy for multi map database");
 
     }
 
-    private MapReadResult readLabels(Tile upperLeft, Tile lowerRight, boolean deduplicate) {
+    private MapReadResult readLabelsPriv(Tile upperLeft, Tile lowerRight) {
         MapReadResult mapReadResult = new MapReadResult();
         boolean isTileFilled = false;
         for (MapDataStore mdb : mapDatabases) {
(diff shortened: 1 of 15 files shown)


mapsforge-map/src/main/java/org/mapsforge/map/datastore/MultiMapDataStore.java
```diff
@@ -152,15 +152,15 @@ public MapReadResult readLabels(Tile tile) {
...
             case DEDUPLICATE:
-                return readLabels(tile, true);
+                return readLabelsPriv(tile).deduplicate();
         }
         throw new IllegalStateException("Invalid data policy for multi map database");
 
     }
 
-    private MapReadResult readLabels(Tile tile, boolean deduplicate) {
+    private MapReadResult readLabelsPriv(Tile tile) {
```
devemux86, line 163: @Sublimis Let's use a more understandable name in these methods.

Either the whole word "private" or better a different name, like `processLabels`, `searchLabels`, etc.

devemux86:
@Sublimis Thanks, it looks good. :+1: 

I merged it in branch [issue_1635](https://github.com/mapsforge/mapsforge/tree/issue_1635) if anyone wants to test it.

devemux86:
@Sublimis We can merge this, but the method names must be improved (see https://github.com/mapsforge/mapsforge/pull/1635#discussion_r1921945552).

Sublimis:
@devemux86 Do you mind if we change `private` to `protected` in `MultiMapDataStore` while at it? It somehow doesn't seem natural to have privates in an open source library, esp. for such an ubiquitous class. Also avoids possible future requests like #1512, "can you please make this non-private, we want to extend"...

devemux86:
@Sublimis No problem, although it introduces the problem that then we cannot easily rename / remove them without breaking user apps.

devemux86:
@Sublimis Thanks for the changes. Is it ready?

Should we also change the `DataPolicy` in the [MultiMapLowResWorld](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MultiMapLowResWorld.java) sample?

Sublimis:
Question, since this hasn't been merged yet: Does the order of the elements in the `MapReadResult` matter, i.e. does the order have to be the same before and after deduplication?
Iff not, then additional improvements can be made here, including bringing back the original hashCode() and equals(), for example, as well as not creating additional list structures.

Sublimis:
> @Sublimis Thanks for the changes. Is it ready?
> 
> Should we also change the `DataPolicy` in the [MultiMapLowResWorld](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MultiMapLowResWorld.java) sample?

Maybe, but since it's just a small world map with no expected duplicates, it's not that important. But it can be done for consistency, of course.

Sublimis:
> Question, since this hasn't been merged yet: Does the order of the elements in the `MapReadResult` matter, i.e. does the order have to be the same before and after deduplication? Iff not, then additional improvements can be made here, including bringing back the original hashCode() and equals(), for example, as well as not creating additional list structures.

It's not a big deal because these hashCode()/equals() are not currently used, but just in case they are needed in the future.

devemux86:
> Maybe, but since it's just a small world map with no expected duplicates, it's not that important. But it can be done for consistency, of course.

@Sublimis Let's change also this, as we currently recommend the deduplicate as default?

devemux86:
> Does the order of the elements in the MapReadResult matter, i.e. does the order have to be the same before and after deduplication?

@Sublimis Was it like this before?

devemux86:
> Is the order of elements in the map file expected to affect the order in which elements are displayed on the screen?

@Sublimis The rendering depends on the render theme. So it should not matter?
If performance improvements can be made because of this, we better do them.

Although better to test all of these to be sure? :slightly_smiling_face:

Sublimis:
Okay, let's try it now. The performance probably won't be much better, maybe a few percent at most, it's more about making hashCode() and equals() future-proof if possible.

devemux86:
@Sublimis Do you mean it's ready or more changes are needed?

Sublimis:
Good news! The order of the elements doesn't seem to matter at all, and even the code is notably more efficient, by more than 10%.
We thought we'd just include those improvements here?

devemux86:
> We thought we'd just include those improvements here?

@Sublimis Let's include them here.

Sublimis:
You can leave this percolating here for another day or two AFAIC. The testers are asleep. 🆗

devemux86:
I updated branch [issue_1635](https://github.com/mapsforge/mapsforge/tree/issue_1635) if anyone wants to test it.

devemux86:
@Sublimis Is it ready, should we merge it?

Sublimis:
> @Sublimis Is it ready, should we merge it?

Yup, ready 👍

Issue #776, opened 2016-01-31
Peculiar behavior of new label layer


Klaus-Tockloth:
The new label layer shows a peculiar behavior ...

That's the result **with** the label layer (only a few house numbers):

![bildschirmfoto 2016-01-30 um 22 18 27](https://cloud.githubusercontent.com/assets/5493128/12700929/d6787a4a-c7f6-11e5-86f5-75dc676e4f29.jpg)

And this is the result **without** the label layer (nearly all (but not all) house numbers are visible):

![bildschirmfoto 2016-01-30 um 22 19 07](https://cloud.githubusercontent.com/assets/5493128/12700926/b22545f6-c7f6-11e5-9d34-8358e88f95a9.jpg)

And this is the situation in OSM:
http://www.openstreetmap.org/#map=19/51.96395/7.63848

![bildschirmfoto 2016-01-31 um 08 48 47](https://cloud.githubusercontent.com/assets/5493128/12700946/8fce48f8-c7f7-11e5-9f55-09e6f6423c8a.jpg)

The problem occurs with Cruiser (1.4.0 - 20160130) and with Atlas (1.2.0-20160130).

Klaus

devemux86:
I can confirm the issue - it can also be reproduced with Samples.

Old `LabelLayer`:
![old](https://cloud.githubusercontent.com/assets/3484020/12701477/17c5231a-c813-11e5-8601-a63f145f6ed8.png)

New `LabelLayer`:
![new](https://cloud.githubusercontent.com/assets/3484020/12701479/1ef4f142-c813-11e5-86f2-c07cd38cecf9.png)

(related to #763)

ludwigb:
I think the reason is the following:

as an optimization we (= I, culpa mea) only store ways if they have a name
to them (otherwise we would essentially need to cache all data). The houses
here have probably no name, but a house number.
The house numbers that do show up are probably points (not on ways, thats
is why they are on entrances and the like).

The fix will be to also store ways if they have a house number, or an
elevation ....

Ludwig

applantation:
I just pushed a fix to dev.

devemux86:
That did the job.

It's worth mentioning that we support:
- In nodes: name, addr:housenumber, ele
- In ways: name, addr:housenumber, ref

ludwigb:
Yes, nodes are always included, ways now only if they have additional
information that is displayed as a label.

eartrumpet:
[...]

PR #1636, opened 2025-01-20
MapDataStoreLabelStore: Return more labels without reading all the data

Sublimis:
Quick performance fix for #1633 until a better solution is found (if needed).

image image

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapDataStore.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapDataStore.java
@@ -322,7 +322,7 @@ public abstract class MapDataStore {
      * @return true if the way should be included in the result set
      */
     public boolean wayAsLabelTagFilter(List<Tag> tags) {
-        return !tags.isEmpty();
+        return false;
     }
 
 }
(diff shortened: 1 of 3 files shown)


Sublimis:
Any suggestions on how to fix the checks?

mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapDataStore.java
```diff
@@ -322,7 +322,7 @@ public MapDataStore setPriority(int priority) {
      * @return true if the way should be included in the result set
      */
     public boolean wayAsLabelTagFilter(List<Tag> tags) {
-        return false;
+        return !tags.isEmpty();
```
devemux86, line 325: @Sublimis This seems to change the default behavior for all implementations.

This hook is intended to be overridden by the required implementations only, if needed (details [here](https://github.com/mapsforge/mapsforge/issues/776#issuecomment-182261899)).

devemux86:
> Any suggestions on how to fix the checks?

@Sublimis Check the error message `Execution failed for task ':mapsforge-map-reader:test'`.

As mentioned above this changes library's default behavior, so it cannot be the default.

devemux86:
See #1637

Sublimis:
The test should probably be adjusted to the fact that labels are not just name tags, but also symbols, etc.

devemux86:
> The test should probably be adjusted to the fact that labels are not just name tags, but also symbols, etc.

Labels were defined as data with names. We can not change easily the default public implementation.
That's why the hook was added. Or we need a new `Selector` to cover more cases.

devemux86:
@Sublimis We can close this?

Sublimis:
> @Sublimis We can close this?

Actually yes, performance will be improved in a different way.

PR #1637, opened 2025-01-20
LabelLayerUsingMapDataStoreMapViewer include ways with tags, fix #1631

### LINKED ISSUE #1631: Label layer with MapDataStoreLabelStore shows less symbols

devemux86:
Fix #1631 via `MapDataStore.wayAsLabelTagFilter` hook.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/labels/MapDataStoreLabelStore.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/labels/MapDataStoreLabelStore.java
@@ -47,7 +47,6 @@ public class MapDataStoreLabelStore implements LabelStore {
 
         this.textScale = textScale;
         this.renderThemeFuture = renderThemeFuture;
-        // TODO (2015): what about way symbols, we have the problem that ways without names but symbols will not be included.
         this.standardRenderer = new StandardRenderer(mapDataStore, graphicFactory, true);
         this.displayModel = displayModel;
     }
@@ -69,7 +68,9 @@ public class MapDataStoreLabelStore implements LabelStore {
             RendererJob rendererJob = new RendererJob(upperLeft, this.standardRenderer.mapDataStore, this.renderThemeFuture, this.displayModel, this.textScale, true, true);
             RenderContext renderContext = new RenderContext(rendererJob, standardRenderer.graphicFactory);
 
-            MapReadResult mapReadResult = standardRenderer.mapDataStore.readMapData(upperLeft, lowerRight);
+            // What about way symbols, we have the problem that ways without names but symbols will not be included.
+            // Possibly fixed by including way symbols with tags via the hook MapDataStore.wayAsLabelTagFilter
+            MapReadResult mapReadResult = standardRenderer.mapDataStore.readLabels(upperLeft, lowerRight);
 
             if (mapReadResult == null) {
                 return new ArrayList<>();
--- a/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingMapDataStoreMapViewer.java
+++ b/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/LabelLayerUsingMapDataStoreMapViewer.java
@@ -15,12 +15,18 @@
  */
 package org.mapsforge.samples.android;
 
+import org.mapsforge.core.model.Tag;
 import org.mapsforge.map.android.graphics.AndroidGraphicFactory;
 import org.mapsforge.map.android.util.AndroidUtil;
+import org.mapsforge.map.datastore.MapDataStore;
 import org.mapsforge.map.layer.labels.LabelLayer;
 import org.mapsforge.map.layer.labels.MapDataStoreLabelStore;
 import org.mapsforge.map.layer.labels.ThreadedLabelLayer;
 import org.mapsforge.map.layer.renderer.TileRendererLayer;
+import org.mapsforge.map.reader.MapFile;
+
+import java.io.File;
+import java.util.List;
 
 /**
  * A map viewer that draws the labels onto a single separate layer. The LabelLayer used in this example
@@ -40,6 +46,17 @@ public class LabelLayerUsingMapDataStoreMapViewer extends DefaultTheme {
 
         // Enable rotation gesture
         mapView.getTouchGestureHandler().setRotationEnabled(true);
+    }
+
+    @Override
+    protected MapDataStore getMapFile() {
+        return new MapFile(new File(getMapFileDirectory(), this.getMapFileName())) {
+            @Override
+            public boolean wayAsLabelTagFilter(List<Tag> tags) {
+                // Include ways with tags to cover the ways without names but symbols
+                return !tags.isEmpty();
+            }
+        };
     }
 
     @Override


Sublimis:
You're merging too quickly imho. 🥲  This is a good attempt, but it's also just a workaround. Think about it: If someone is using `MapDataStore.readLabels()` (a public interface), they want to get everything that is considered a "label" by the library, not just name tags. The definition of "label" is very broad and now includes much more than just name tags. This is something that should be improved at the library level, not delegated to users, right?

devemux86:
> using MapDataStore.readLabels() (a public interface), they want to get everything that is considered a "label" by the library

Reading the documentation of the library, it is clear what to expect:

- [Selector documentation](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java#L1104): _"LABELS: poi data and ways that have a name"_
- [readLabels documentation](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/datastore/MapDataStore.java#L139): _"Reads only labels for tile. Labels are pois as well as ways that carry a name tag."_

This is how this selector and readLabels works by default, i.e. with named map elements.
The problem is the implementation of the specific label layer. The selector works as it was designed.

Sublimis:
If `readLabels()` works as expected, and `MapDataStoreLabelStore` uses `readLabels()`, than `MapDataStoreLabelStore` also works as expected. Do you see the problem? At the very least, we have to agree that the definition of "label" is not the same between `MapDataStoreLabelStore` and `MapDataStore.readLabels()`, and that is bad by itself.

devemux86:
Only the label layer with the `MapDataStoreLabelStore` has the problem.
Without label layer or with default label layer everything works correctly.

That's why I mentioned that the improvement is not about changing the default public behavior of a library selector.
Either improve the implementation in the problematic class `MapDataStoreLabelStore` or use a new selector there.

devemux86:
We should not forget that Mapsforge is used by many applications. :slightly_smiling_face: 
It's one thing to change/remove a public method for a better implementation.
And another thing to change (internally) the public behavior of an API to a different design.

Sublimis:
But `readLabels()` *does not work correctly* for the past 10 years, let's fix it! 😃   It's probably just a matter of no one noticing / caring enough. `EncodingTest` itself hasn't been changed in the last 10 years, and the poor thing is probably still working with an outdated label definition. Or if you want to keep this broken behavior we can introduce the new and fixed `readLabelsEx()` 🙂 

When the new separate label layer was introduced a long time ago, the label definition was basically quietly changed, this should never have happened that way.

Let's recall one comment from that time, that best defines a "new" label: 

> Those labels should not be on the tile layer either, a main motivation for
the LabelLayer is support rotation, so everything that needs rotation
should go onto the LabelLayer.

This just needs to be placed on a solid foundation within the library.
Source: https://github.com/mapsforge/mapsforge/issues/776#issuecomment-181144972

> Without label layer or with default label layer everything works correctly.

Without label layer we have nothing to talk about, but are you sure about the second part? What is a "default label layer"? `MapFile` overrides the `MapDataStore.readLabels()`, and from there on it works incorrectly (per "new" label definition). The default `MapDataStore.readLabels()` (the one that reads all map data instead) is never called anywhere within the library.

devemux86:
Let's not confuse the "label" word:
- Label layer displays both labels and symbols.
- Selector.LABELS and readLabels are about elements with names.

What will happen with users / apps that want to get just named elements,
if we suddenly and silently return all elements with tags?

If the label layer with `MapDataStoreLabelStore` does not work correctly, indeed let's fix it.
With extra selector, extra code, something that will not break all user apps.

devemux86:
> MapFile overrides the MapDataStore.readLabels(), and from there on it works incorrectly (per "new" label definition).

Why? It correctly uses the `Selector.LABELS`.

> The default MapDataStore.readLabels() (the one that reads all map data instead) is never called anywhere within the library.

Its a public method, it may be used by apps. And it works according to its documentation.

e.g. [ReverseGeocodeViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/ReverseGeocodeViewer.java) sample

---

Let's add a new selector (working with elements with tags) and use it in `MapDataStoreLabelStore`.

Sublimis:
Okay. So basically we have two separate concepts sharing the same word. This is then also something that should be improved within the code/javadoc, if we really want to leave the exact same word for both. Maybe it would be better to use some other word not used elsewhere? Mark, stamp, flag, token, sign, extract, ...?

devemux86:
If there is need, let's improve documentation.
Public API exists for many years, we cannot change it now.

Sublimis:
> > MapFile overrides the MapDataStore.readLabels(), and from there on it works incorrectly (per "new" label definition).
> 
> Why? It correctly uses the `Selector.LABELS`.

> per "new" label definition

Sublimis:
> Let's add a new selector (working with elements with tags) and use it in `MapDataStoreLabelStore`.

Sounds good.

Sublimis:
> > Let's add a new selector (working with elements with tags) and use it in `MapDataStoreLabelStore`.
> 
> Sounds good.

Just a few questions, what elements are expected to have tags? What are tags used for?
If, let's imagine, in the future all map elements are to have at least one tag, then that will be the same as `Selector.ALL`, and we are trying to avoid that here for performance reasons.

devemux86:
> per "new" label definition

There is no new label definition. A new selector maybe, that could do different things.
There are already docs in these methods and they seem to work according to it.

The only use of readLabels inside the repository is in ReverseGeocodeViewer sample where its doc mentions:
_"Read all **labeled** POI and ways for the area covered by the tiles under touch"_

devemux86:
About your question, the only problem exists in the `MapDataStoreLabelStore`.
We do not need anywhere else an extra selector.

Don't all elements have tags? Otherwise how we could use them in themes.

Sublimis:
> > per "new" label definition
> 
> There is no new label definition. A new selector maybe, that could do different things. There are already docs in these methods and they seem to work according to it.
> 
> The only use of readLabels inside the repository is in ReverseGeocodeViewer sample where its doc mentions: _"Read all **labeled** POI and ways for the area covered by the tiles under touch"_

Just to be clear, these are our ad-hoc definitions used to distinguish two separate concepts that share the same word as mentioned above:

* "New" label definition: Everything that needs rotation.
* "Old" label definition: Element with a name.

Sublimis:
> About your question, the only problem exists in the `MapDataStoreLabelStore`. We do not need anywhere else an extra selector.

You cannot possibly know that. Anyone using the public `readLabels()` method sees the same data, the only difference is that there are no us in their code to tell them if what they're seeing is wrong or right 🙂 

> Don't all elements have tags? Otherwise how we could use them in themes.

Really don't know. It appears not, because the previous workaround we shared in PR (the `!tags.isEmpty()` one) showed immediate performance improvement, that could not be possible if all elements had tags...

PR #1640, opened 2025-01-21
Avoid confusion with the word "labels" in context of a map file or store

Sublimis:
Avoid confusion with the word "labels" in context of a map file or store - an elegant solution.

* No behavior change, everything continues to work as before.
* Minimizes the number of changes in the code.
* Users are notified that two (rarely used) methods have been renamed.

#1637 
#1636

(Note: MultiMapDataStore adaptation is left for a separate step, to avoid conflicts with #1635 )


devemux86:
@Sublimis Thanks, although it would be better not to rename the public API.

If we decide to rename it, we should just rename it.
Without also have and maintain the old API, as it will be mentioned in the changelog.

> (Note: MultiMapDataStore adaptation is left for a separate step, to avoid conflicts with https://github.com/mapsforge/mapsforge/pull/1635 )

If we go for it, the pull request must be complete in all related classes.
#1635 can be adapted later.

devemux86:
@Sublimis It's better to rename / remove them in all classes that they exist.
Instead of maintaining deprecated code. Changelog exists for this purpose.

devemux86:
> (Note: MultiMapDataStore adaptation is left for a separate step, to avoid conflicts with https://github.com/mapsforge/mapsforge/pull/1635 )

@Sublimis Now that #1635 was merged, this needs to rename the `readLabels` in all classes.

Also please remove the deprecated methods and enum. We can mention them in changelog.

devemux86:
@Sublimis Thanks!

PR #1641, opened 2025-01-23
Smoother and more accurate paths with floating point coordinates

Sublimis:
After/before animation:

2025-01-23_smooth_path


devemux86:
@Sublimis Thanks!

PR #1643, opened 2025-01-24
Android: Use cache folder/subfolder to cache SVG files

Sublimis:
On Android there is an elegant concept of a "cache folder" which should be used to cache SVG files. No application wants thousands of files in the root of its private folder, which is currently done. Even in the cache folder we create a separate subfolder for that.

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
@@ -31,6 +31,7 @@ import org.mapsforge.core.mapelements.PointTextContainer;
 import org.mapsforge.core.mapelements.SymbolContainer;
 import org.mapsforge.core.model.BoundingBox;
 import org.mapsforge.core.model.Point;
+import org.mapsforge.core.util.Constants;
 import org.mapsforge.map.model.DisplayModel;
 
 import java.io.*;
@@ -63,7 +64,6 @@ public final class AndroidGraphicFactory implements GraphicFactory {
 
     public static final Config MONO_ALPHA_BITMAP = Config.ALPHA_8;
 
-    public final String LibraryFolderName = "mapsforge";
     private final Object cacheFolderSync = new Object();
     private volatile File cacheFolder = null;
 
@@ -332,7 +332,7 @@ public final class AndroidGraphicFactory implements GraphicFactory {
             synchronized (this.cacheFolderSync) {
                 ourCacheFolder = this.cacheFolder;
                 if (ourCacheFolder == null) {
-                    ourCacheFolder = new File(this.context.getCacheDir(), LibraryFolderName);
+                    ourCacheFolder = new File(this.context.getCacheDir(), Constants.LIBRARY_FILE_NAME);
                     ourCacheFolder.mkdirs();
                     this.cacheFolder = ourCacheFolder;
                 }
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
@@ -31,7 +31,6 @@ import org.mapsforge.core.mapelements.PointTextContainer;
 import org.mapsforge.core.mapelements.SymbolContainer;
 import org.mapsforge.core.model.BoundingBox;
 import org.mapsforge.core.model.Point;
-import org.mapsforge.core.util.Constants;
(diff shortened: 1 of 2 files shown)


mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
```diff
@@ -62,6 +63,10 @@ public final class AndroidGraphicFactory implements GraphicFactory {
 
     public static final Config MONO_ALPHA_BITMAP = Config.ALPHA_8;
 
+    public final String LibraryFolderName = "mapsforge";
```
devemux86, line 66: @Sublimis Please use a variable that starts with lowercase letters.

devemux86:
@Sublimis Note that there is the `setSvgCacheDir` method to customize the svg cache folder.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
Sublimis: Hm, this is not a variable but a (non-static) constant...?
In fact, maybe this should be put into the Constants class in some way.

Sublimis:
> @Sublimis Note that there is the `setSvgCacheDir` method to customize the svg cache folder.

That doesn't affect what was done here at all.

mapsforge-map-android/src/main/java/org/mapsforge/map/android/graphics/AndroidGraphicFactory.java
Sublimis: What's the final suggestion then? Make it static, uppercase, and include in Constants, or something else?

Sublimis: Are you sure there wouldn't be any benefit to putting this in Constants? Surely somewhere else this could be used or will become necessary at some point (if it hasn't already), for example FileSystemTileCache could also benefit from this.

Sublimis: When the time comes for someone to use such a constant, no one will remember that there is already one buried in the android svg cache 🥲

devemux86: @Sublimis Does it matter?

It should not be a constant, but a variable, as other users may prefer a different folder name.
Apps may want to store there svg, png, tiles, etc. Each app may want different folder names.

devemux86: > Also note the unfortunate naming of methods there. E.g. openFileInput() uses the folder set by setSvgCacheDir(), so it's not a general-purpose method.

Mapsforge has a very long history of too many years and many developers.
It is difficult to rename the whole public API without serious reason. :slightly_smiling_face:

devemux86:
> is there still a purpose for this folder?

It contains the collection if the original svg icons.
_Some of them_ were processed and included in the themes. Original icons may be useful in the future.

devemux86:
@Sublimis My only comment here is to rename the constant, then we can merge it.

Sublimis:
Two commits, the previous one is just to show you what we had in mind:

https://github.com/mapsforge/mapsforge/blob/17dda95d84aefeecaaa501de4597a7fb78dbae5a/mapsforge-core/src/main/java/org/mapsforge/core/util/Constants.java

```java
package org.mapsforge.core.util;

public final class Constants {

    /**
     * Display-friendly, human readable name of the library.
     */
    public static final String LIBRARY_DISPLAY_NAME = "Mapsforge";

    /**
     * Filename-friendly, machine readable name of the library.
     * To be used when there's a need to name a file, folder, or similar.
     */
    public static final String LIBRARY_FILE_NAME = "mapsforge";

    /**
     * Maximum amount of memory that the Java virtual machine will attempt to use, in megabytes.
     */
    public static final long MAX_MEMORY_MB = Runtime.getRuntime().maxMemory() / 1000 / 1000;

    private Constants() {
        throw new IllegalStateException();
    }
}
```

It's a little strange that the library doesn't "know" its name all these years 🙂

devemux86:
@Sublimis There is no perfect library! I implemented **map rotation** only recently, after so many years.

If it is so important, we can move the constants in the `Constants` class.

Sublimis:
It's not important, it's just practical and aesthetic 🙂 We can leave this for the future if it's easier that way.

devemux86:
@Sublimis Thanks!

PR #1644, opened 2025-01-27
Fix: Elevation contours should not be drawn above roads

Sublimis:
Fix: Elevation contours should not be drawn above roads and everything in Motorider themes (only above hill shading)

#1641 as an example

### DIFF
--- a/mapsforge-themes/src/main/resources/assets/mapsforge/motorider-dark.xml
+++ b/mapsforge-themes/src/main/resources/assets/mapsforge/motorider-dark.xml
@@ -704,6 +704,72 @@
     <!-- hillshading -->
     <hillshading color="#FFEEEEEE" zoom-min="9" zoom-max="17" />
 
+    <!-- contour lines LAYER, to work with Freizeitkarte and OpenAndroMaps -->
+    <rule cat="contour" e="any" k="contour_ext" v="*">
+        <rule e="way" k="*" v="*" zoom-max="12">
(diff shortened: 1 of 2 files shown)


devemux86:
@Sublimis Where [OpenAndroMaps Elevate theme](https://www.openandromaps.org/en/legend/elevate-mountain-hike-theme) shows the contours?

cc @MotoUKRider

devemux86:
@MotoUKRider @Sublimis 

OpenAndroMaps Elevate seems to draw the contours below the roads but above all other polygons.
So we should probably use a similar rendering.

devemux86:
@MotoUKRider Labels cannot be hidden, they are rendered in a separate layer on top.

It's better to not complicate things. :slightly_smiling_face: 
Just check how [Elevate theme](https://www.openandromaps.org/en/legend/elevate-mountain-hike-theme) works in all apps.

Cruiser

devemux86:
@MotoUKRider @Sublimis 

I merged the changes in https://github.com/mapsforge/mapsforge/pull/1645.

devemux86:
@MotoUKRider @Sublimis Thanks for the improvements!

devemux86:
@Sublimis @MotoUKRider Thank you both for the improvements!

Sublimis:
Elevate is doing similar, except they seem to draw contours above tunnels only 🙂
So the order is this:

```xml
<hillshading/>
...
<!-- tunnels -->
...
<!-- contour lines -->
...
```

PR #1655, opened 2025-02-17
Android: option for non-transparent bitmaps to use ARGB_8888 instead of RGB_565

devemux86:


Sublimis:
After everything said in #1654, why is 16-bit color still the default? Analysis there wasn't detailed enough lol?

It's better to just let go of 16-bit color. No parameter/option.
If you really need 16-bit color for some niche reason, it's better to implement a dedicated `create16bitBitmap()` method for external users, and let the rest of us enjoy true color tiles out of the box.

16-bit color should be regarded as having special usage in (micro-)optimizations, as a last resort when dealing with tons of large bitmaps at the same time (think gallery or photoshop app), and even then it's a big if. This library is nowhere near such reasons.

If you need a little more convincing, look at hill shading gradients in 16-bit color...
Is there a need for additional analysis of the pros and cons?

devemux86:
@Sublimis This was already discussed. :slightly_smiling_face: 

We did not see significant improvements, unless using hillshading (which is not used by all users)
and if we zoom in and check very carefully the screenshots.

So the current config has no significant reason to change as default affecting all apps out there.
At least not now, suddenly without testing by more users.

This is the purpose of the parameters, everyone enables whatever they want for their apps.

Sublimis:
Understanding your reasons, not to say fears, we see it differently:
You've been forcing 16-bit color to all apps for the last 10 years (at least).

Now it's not forced any more, but it's still the default parameter that no one will touch. It's a good first step, better late than never, but still.

Have you considered the fact that there are other render themes out there besides Motorider?

devemux86:
Sorry, I do not understand the question.

Same color values in render themes produce the same color result on the screen.

Sublimis:
> Also let's use correct boolean in #1656 to activate new things, instead of the previous confusing reverse boolean behavior.

But it won't be reverse one day 😅 (the sooner the better!)

Sublimis:
> Same color values in render themes produce the same color result on the screen.

There it is!
Misconception about color spaces.

The above statement is incorrect.

devemux86:
So the `#0000FF` can produce different blue color in different themes?

Sublimis:
> So the `#0000FF` can produce different blue color in different themes?

Yes.

https://en.wikipedia.org/wiki/Indexed_color
https://en.wikipedia.org/wiki/High_color

devemux86:
Sure, but that's why there are options to keep everyone happy. :slightly_smiling_face: 

It's up to the developer to choose what they need,
32bit or possibly less memory usage.

Sublimis:
In short, the exact colors that will be seen on the screen depend entirely on the color palette used in the 16-bit bitmap (i.e. the rendering theme). There is no 1-to-1 color mapping, only approximations. Some colors will get their exact 24-bit equivalent, but how many will not depends entirely on the properties of the color palette used and the relationships between the colors.

If you're now thinking "but the rendering theme doesn't use that many colors!", let's not forget that we have **antialiasing enabled**.

Sublimis:
> Sure, but that's why there are options to keep everyone happy. 🙂

Now to be cynical, where has the option been for the last 10 years? 😅

> It's up to the developer to choose what they need, 32bit or possibly less memory usage.

Have you done the math? How much memory is saved?

devemux86:
We have to keep in mind that Mapsforge is not like [VTM](https://github.com/mapsforge/vtm) which uses OpenGL and GPU natively.
It already uses a lot of memory, it's nice to have options like this that help in all possible ways.

Sublimis:
Understandable, but as already stated, since year 2017 bitmaps are stored in the **native heap**. Native heap uses **all the memory the device has**. The year is 2025, devices have 4GB+ of RAM, and we're talking 1-2MB savings here. The 16-bit default is now just on the wrong side of history.

PR #1657, opened 2025-02-19
Fix: libhwui.so "null pointer dereference" SIGSEGV (Android)

Sublimis:
Another "not a new feature". We've been using it in Urban Biker for a while now.
Take it or leave it 😁

### DIFF
--- a/mapsforge-map-android/src/main/java/org/mapsforge/map/android/util/AndroidParameters.java
+++ b/mapsforge-map-android/src/main/java/org/mapsforge/map/android/util/AndroidParameters.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2025 Sublimis
+ *
+ * This program is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free Software
+ * Foundation, either version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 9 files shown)


mapsforge-map-android/src/main/java/org/mapsforge/map/android/util/AndroidParameters.java
```diff
@@ -0,0 +1,27 @@
...
+import android.os.Build;
+
+public final class AndroidParameters {
+
+    /**
+     * To prevent {@code libhwui.so} "null pointer dereference" SIGSEGV crashes on some Androids.
+     * The only drawback seems to be that line joins of a few paths may not be rounded.
+     * Defaults to {@code true} for Android API levels 32 and below.
+     */
+    public static boolean ANDROID_LIBHWUI_FIX = Build.VERSION.SDK_INT <= Build.VERSION_CODES.S_V2;
```
devemux86, line 26: @Sublimis Why this particular Android version?

On which Android versions have you seen this problem happen?

Sublimis: Older obviously. Some change happened between Android 12 and 13 that most probably fixes the bug. Nonetheless, there may also be performance benefits, as drawing a bunch of lines is faster than using a path from our earlier experience.

PR #1658, opened 2025-02-21
AndroidParameters constructor, #1657

devemux86:


Sublimis:
This doesn't provide any benefit. It would be best to completely overhaul the parameter classes while at it; perhaps make a distinction between those parameters that are okay to change at runtime and those that should not be changed after setup. The former **should be marked as volatile**, the latter should be implemented as part of a factory that would create a solid immutable object.

devemux86:
Just being the same as other similar classes. :slightly_smiling_face:

Sublimis:
Hmm, sometimes it seems like we don't speak the same language 🙂 Yes, this concerns both `Parameters` (primarily) and `AndroidParameters` classes**es**. Are there any others?

It is not _guaranteed_ that these parameter fields will be synchronized in any way. When someone decides to update one of the parameters during (multi-threaded) execution, they could run into troubles. That was just one of the many reasons why we asked you nicely to set 32-bit color as the default or, for example, not to use the parameter at all.

devemux86:
Another example is the various `Constants` classes.

---

`Parameters` is to be set once before initializing and using the library.
_Its purpose was never to be changed during the library usage._

e.g. desktop [Samples](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-awt/src/main/java/org/mapsforge/samples/awt/Samples.java)

Sublimis:
Read this to better understand the implications of not using `volatile` for such fields: https://stackoverflow.com/questions/2423622/volatile-vs-static-in-java

In particular:

> Even if you access a static value through multiple threads, each thread can have its local cached copy!

And:

> Volatile fields are special fields which are used for communicating state between threads. Each read of a volatile will see the last write to that volatile by any thread; in effect, they are designated by the programmer as fields for which it is never acceptable to see a "stale" value as a result of caching or reordering. The compiler and runtime are prohibited from allocating them in registers. They must also ensure that after they are written, they are flushed out of the cache to main memory, so they can immediately become visible to other threads. Similarly, before a volatile field is read, the cache must be invalidated so that the value in main memory, not the local processor cache, is the one seen.

from https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html#volatile.

Earlier we guesstimated that probably no one messes with the Parameters class in production and large-scale applications, because otherwise we'd probably see some reports of problems caused by this.

Issue #1659, opened 2025-02-24
world.map doesn't work on zoom level 22-24


mg4gh:
When you just load the world.map from mapsforge maps and zoom to level 22-24, then it isn't displayed anymore correctly. The problem seems to be the calculation of the insideArea in the DatabaseRenderer and the call to `renderContext.canvasRasterer.fillOutsideAreas(Color.TRANSPARENT, insideArea);`. In the method fillOutsideAreas there is a cast to int, which may give wrong result if the map has the bounding box of the whole world.

So obviously nobody will just visualize the world map and go to this zoom level, but with nice feature of the MultiMapDataStore it make sense to have the world map and e.g. the germany.map from openandromaps loaded. If you zoom to 22 or larger, then the visualization of the germany.map doesn't work anymore due to this problem.

A user of my app has this problem reported.

devemux86:
@mg4gh Thanks for the report and also for checking the code. 👍 

This should be fixed now with https://github.com/mapsforge/mapsforge/pull/1660.

Unless using indoor maps, there is no reason to use such high zoom levels above 20-22.

You can control the zoom level limits with:
```java
mapView.getModel().mapViewPosition.setZoomLevelMin(2);
mapView.getModel().mapViewPosition.setZoomLevelMax(20);
```

Zoom level 20 has 0.1 m / pixel: https://wiki.openstreetmap.org/wiki/Zoom_levels

mg4gh:
Thanks for the fast reply. I've just integrated and tested and ... surprisingly the problem still exists. I wonder whether the native code is doing some cast to integer too.
 
I know that I can easily limit the zoom level - but in rare cases (problems with my routing) it was very useful to zoom in and see each single microdegree of the way objects, so I rather would like to keep this in.

I wonder what is the use case for this clip operation. A map like germany anyway does not fill a rectangular bouning box, where I can clip it in a way that make sense. And with a MultiMapDataStore which contains also the world.map there isn't anything to clip. So what's the rationale to do this operation?

mg4gh:
Android 15 on an emulator.

Do you expect a different behaviour with a real device?

devemux86:
There was floating point overflow, at least on Android whose methods work with `float`.

[MultiMapLowResWorld](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MultiMapLowResWorld.java) shows 2 ways to limit the zoom levels of the map files.
[...]

PR #1663, opened 2025-02-26
Fix: Incorrect use of Double.MIN_VALUE

Sublimis:
Found one more mistake with floating point MIN_VALUE like the one mentioned in #1659


devemux86:
@Sublimis Thanks!

We could also use the `POSITIVE_INFINITY`, `NEGATIVE_INFINITY` that is used elsewhere.

Sublimis:
> We could also use the `POSITIVE_INFINITY`, `NEGATIVE_INFINITY` that is used elsewhere.

That's true, and probably would not have led to a mistake if it had been used originally, but works just as well like this

PR #1664, opened 2025-02-26
* Fix: Normalize input to CanvasRasterer.fillOutsideAreas()

Sublimis:
* Fix: Normalize input to `CanvasRasterer.fillOutsideAreas()` to prevent overwhelming the canvas with extreme coordinate values

#1659

### DIFF
--- a/mapsforge-core/src/main/java/org/mapsforge/core/model/Rectangle.java
+++ b/mapsforge-core/src/main/java/org/mapsforge/core/model/Rectangle.java
@@ -218,15 +218,15 @@ public class Rectangle implements Serializable {
         return new Rectangle(this.left + origin.x, this.top + origin.y, this.right + origin.x, this.bottom + origin.y);
     }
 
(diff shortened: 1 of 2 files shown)


Sublimis:
Take 2.

Now, the most elegant solution to all of this would be to simply clear the `MapView` with `map-background-outside` color before drawing, and not delegate that to the renderer, isn't it so? In that scenario, this should only be considered a temporary solution...

devemux86:
@Sublimis Thanks, this seems to work much better.

Sublimis:
> the most elegant solution to all of this would be to simply clear the `MapView` with `map-background-outside` color before drawing, and not delegate that to the renderer

It works, although it's probably not as general solution as it could be, so what about this?
This seems easily implementable by providing the `map-background-outside` value to this line:

https://github.com/mapsforge/mapsforge/blob/b358a273140728885f6a8fc9756b677a45990cfb/mapsforge-map/src/main/java/org/mapsforge/map/layer/TileLayer.java#L91

PR #1711, opened 2025-07-02
Disable Maven publish because OSSRH service reached end-of-life on 2025-06-30

devemux86:
OSSRH service reached end-of-life on 2025-06-30

https://central.sonatype.org/news/20250326_ossrh_sunset/

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -24,9 +24,3 @@ jobs:
       run: chmod +x gradlew
     - name: Build with Gradle
       run: ./gradlew build
-    - name: Publish
-      if: github.event_name != 'pull_request'
-      env:
-        SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
-        SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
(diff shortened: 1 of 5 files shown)


Sublimis:
Well, you shouldn't just give up on Maven so boldly for such a weak reason!
https://central.sonatype.org/pages/ossrh-eol/#publishing-to-central-portal

devemux86:
They once again require developers to change publishing workflows.

And they made the changes without providing an official Gradle plugin:
https://central.sonatype.org/publish/publish-portal-gradle/

There is simply no free time to retest Maven changes for the umpteenth time.

If anyone has free time to experiment with Maven, feel free to do so.

Sublimis:
Hopefully (just guessing) this is only a temporary hiccup, caused by increased security, until a new guide or tools come up. There is no viable alternative, after all.
Uploading manually isn't an option? New versions come out quite infrequently.

Sublimis:
https://committing-crimes.com/articles/2024-09-09-jitpack/

Let's just leave this here (again) for future reference.

It's a shame that Maven makes publishing so difficult and we hope they come to their senses, but yes, if someone doesn't want to use JitPack (as we don't) they can always compile from source (as we do).

hannesa2:
If someone comes with a working blueprint how to make it work again: this very welcome.
But I don't waste my time here (again)

PR #1717, opened 2025-07-25
New Display.FORCED enum and behavior

Sublimis:
* New Display.FORCED enum and behavior

Regarding https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-3046372969

### DIFF
--- a/docs/Rendertheme.md
+++ b/docs/Rendertheme.md
@@ -274,7 +274,7 @@ Labels and icons are now drawn in order of priority, higher priorities first. Th
 
 ## Display
 The display directive has been added wherever priorities can be used. The following values can be used:
-- ifspace: An element is displayed if there is space for it and no higher priority element will take it. Recommended.
+- ifspace: An element is displayed if there is space for it and no higher priority element will take it. This is the default.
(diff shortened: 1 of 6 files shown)


Sublimis:
Updated

devemux86:
@Sublimis Thanks for working on this.

I made a quick test, changing the fuel and charging stations to `forced`.
While the stations are visible on zoom 13, they are not visible on zoom 11.

I don't know if this is related to the priorities in the theme or more work is needed?
Should we expect the `forced` to be visible on all zoom levels regardless of priority?

Sublimis:
> I made a quick test, changing the fuel and charging stations to `forced`. While the stations are visible on zoom 13, they are not visible on zoom 11.
> 
> I don't know if this is related to the priorities in the theme or more work is needed? Should we expect the `forced` to be visible on all zoom levels regardless of priority?

There was an overridden method in `PointTextContainer` that was forgotten, now fixed, that might change the behavior in your test.

Not sure about zoom levels and priorities. It seems unrealistic that the icon would be visible at *all* zoom levels. Probably the reader just doesn't load the icon if the zoom-min doesn't match (so the icon never even gets to the element placing / collision detection algorithm).

devemux86:
@voldapet @menion Can you check this pull request?

devemux86:
@Sublimis Thanks!

PR #1726, opened 2025-09-06
Use a general xs:string for pathText and caption keys instead of fixed sets

tumic0:
OpenAndroMaps maps and themes use aditional keys like "ref_hike" or "ref_cycle" that can not be used with valid render themes. The Mapsforge library and even the alternative implementation in GPXSee can already handle arbitrary keys, so extending the theme should cause no harm to any user while at the same time making themes like Elevate valid.

### DIFF
--- a/resources/renderTheme.xsd
+++ b/resources/renderTheme.xsd
@@ -131,15 +131,6 @@
         </xs:restriction>
     </xs:simpleType>
 
-    <xs:simpleType name="textKey">
-        <xs:restriction base="xs:string">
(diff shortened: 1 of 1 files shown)


devemux86:
Note: xml schema is only about the theme validation in the xml editors.

It does not change how the OpenAndroMaps and other maps and themes can be used.

So if additional keys like "ref_hike" or "ref_cycle" are already displayed if included in map files and themes,
then we can merge this.

PR #1729, opened 2025-09-12
DirectRenderer changes

menion:
Best flow for using `DirectRenderer` seems to be now:

- using `TileBasedLabelStore` to get to labels for already loaded tiles. I've previously used `MapDataStoreLabelStore`, but it caused huge problems in low zoom levels (loading of too much data)
- rendering of labels over `LabelLayer` 

We need to process labels, but not render them. So they are stored in `TileBasedLabelStore`. In the end, it is only needed to modify `DirectRenderer` to enable caching (optional) + using generic `LabelStore`.

### DIFF
--- a/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DirectRenderer.java
+++ b/mapsforge-map/src/main/java/org/mapsforge/map/layer/renderer/DirectRenderer.java
@@ -42,6 +42,7 @@ public class DirectRenderer extends DatabaseRenderer {
      * @param graphicFactory    the graphic factory.
      * @param labelStore        from where labels are read.
      * @param renderLabels      if labels should be rendered.
+     * @param cacheLabels       if labels should be cached.
      * @param hillsRenderConfig the hillshading setup to be used (can be null).
      */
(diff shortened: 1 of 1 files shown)


devemux86:
`DirectRenderer` now has the correct constructor with `LabelStore` and `cacheLabels` options.
Users can now use freely all the options they want. The previous constructor was very limited.

---

If want to examine "public" APIs, better see how `Display.ALWAYS` broke many themes and apps:
- https://www.openandromaps.org/en/oam-forums/topic/osmc-beacon-icons
- https://github.com/mapsforge/mapsforge/discussions/1731
- https://github.com/mapsforge/mapsforge/pull/1586#issuecomment-3046372969

Asking everyone to change a theme rule, that many apps do not currently support, is not nice.

Sublimis:
@devemux86 

Come on, man. Testing went on for a while, reviewing went on and no one complained for almost a year about such a small thing. And you've only now decided to throw it in our faces? After our massive contributions have finally, after so many years, elevated this library to a level that is not embarrassing?

To get back to the point, the primary utility of `DirectRenderer` is what it says to clients: There is a supported way by which you can get fully rendered tile bitmaps and do whatever you want with them.

This feature should be embraced, simplified, and advertised much more. It's a trivial, elementary, and expected feature, and almost all mapping SDKs have it. Yet, for some reason, you seem to be persistently (since always) undermining attempts to make it happen, making apps that use this look like they're hacking something. Why?

PR #1740, opened 2025-09-19
- improved getting correct OSMTags

menion:
This commit refactors the `OSMTagMapping` class to simplify and clarify the logic for retrieving `OSMTag` objects.

**Explanation**:
We did this change with @voldapet somewhere in March 2025.
The reason: (not 100% sure now) was because of incorrect mapping `int`/`short` values of Netherlands/Belgium crossings in the `rcn_ref` and `rwn_ref` tags. We have already used this code in our generator for a few months and it should work as expected. Unfortunately, a precise check of the logic, mainly in the `OSMTagMapping.getTag` method, is highly recommended.

**Key changes**:
-   Created a `getTag` private helper method to consolidate the common logic for retrieving tags from `stringToPoiTag` and `stringToWayTag` maps. This method now uses the `ValueType` enum for clearer type checking.
-   Updated `getPoiTag` and `getWayTag` to utilise the new `getTag` method and the `GetTagAlternative` interface.

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/OSMTagMapping.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/OSMTagMapping.java
@@ -127,6 +127,20 @@ public final class OSMTagMapping {
         }
     }
 
+    /**
+     * Type of the variable stored in the key/value pair.
+     */
+    private enum ValueType {
+        /**
+         * Number value. May be "byte", "short", "integer", "float".
+         */
+        NUMBER,
+        /**
+         * Plain text value (also used as fallback).
+         */
+        TEXT
+    }
+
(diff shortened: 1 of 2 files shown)


devemux86:
@menion It seems that now the included data is changing in the wrong way?

The purpose of the PR is to improve some methods?
But we should not change how the data are parsed.

If we compare berlin.map with `master` and this PR we get different rendering.
The map rendering and the included data should remain the same?

Even the map size is changed from 48.9 MB to 49.7 MB.

Now there is a tall building in 2nd image and not the detailed smaller buildings in 1st image.
And the images now have different colors.

The comparison is made with [VTM](https://github.com/mapsforge/vtm), checking the 3D buildings and their colors:

---

**Master**
master

---

**Pull request**
pull_request

PR #1741, opened 2025-09-19
Map writer: optimize geometry processing and clipping

menion:
Based on my attempt to speed up map generation, here are two improvements. It has a significant effect when generating over "RAM" tile cache. When using HDD, this does not have significant effect, as cliping itself consume most of time.

Again, this is used in our production generator for a few months and no negative effect detected. In the case of RAM generation, we reduced time by approx 30%!

Major changes:
- **Dependencies update:** updated Guava and JTS library
- **Pre-calculate tile geometry:** The JTS `Geometry` representation of a tile is now calculated once and reused for all ways within that tile, speed up and reducing redundant calculations.
- **Optimized clipping:**
    - The `clipToTile` method now accepts a pre-calculated tile `Geometry` instead of a `TileCoordinate` and enlargement factor, simplifying the call site.
    - An early exit is added to `clipToTile` if the geometry is already fully contained within the tile's envelope, avoiding unnecessary intersection calculations.
- **Use `StringBuilder` for preferred languages:** Improved string concatenation performance when constructing the preferred languages string.

### DIFF
--- a/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
+++ b/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/MapFileWriter.java
@@ -90,11 +90,12 @@ public final class MapFileWriter {
             if (geometry == null) {
                 way.setInvalid(true);
                 throw new Exception("cannot create geometry for way with id: " + way.getId());
-            } else if (!geometry.isValid()) {
+            }
+            if (!geometry.isValid()) {
                 geometry = JTSUtils.repairInvalidPolygon(geometry);
                 if (!geometry.isValid()) {
-                    LOGGER.warning("load(), invalid geometry even after attempt to fix " + way.getId());
-                    return null;
+                    way.setInvalid(true);
+                    throw new Exception("invalid geometry even after attempt to fix " + way.getId());
                 }
             }
             return geometry;


mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/GeoUtils.java
```diff
@@ -77,35 +77,24 @@ public final class GeoUtils {
...
-                geometry = JTSUtils.repairInvalidPolygon(geometry);
-                if (!geometry.isValid()) {
-                    LOGGER.warning("invalid geometry even after attempt to fix " + way.getId());
-                }
+            if (tileAsGeometry.getEnvelopeInternal().contains(geometry.getEnvelopeInternal())) {
+                ret = geometry;
+            } else {
+                ret = tileAsGeometry.intersection(geometry);
             }
-            ret = tileBBJTS.intersection(geometry);
```
devemux86, line 105: @menion The rendering problem seems to be caused here:
- Declare `ret` only as intersection like before and the rendering seems to work correctly.

- Also, why remove the repair section if geometry is not valid?
```java
if (!geometry.isValid()) {
    // this should stop the problem of non-noded intersections that trigger an error when
    // clipping
    LOGGER.warning("invalid geometry prior to tile clipping, trying to repair " + way.getId());
    geometry = JTSUtils.repairInvalidPolygon(geometry);
    if (!geometry.isValid()) {
        LOGGER.warning("invalid geometry even after attempt to fix " + way.getId());
    }
}
ret = tileAsGeometry.intersection(geometry);
```

devemux86:
@menion Thanks!

This is a very nice performance improvement, so we should not forget it.

With the changes suggested [above](https://github.com/mapsforge/mapsforge/pull/1741#discussion_r2375578442), current tests seem to work correctly.

devemux86:
@menion 

> and Kotlin of course 😊

I prefer not to introduce a new language into the project. :slightly_smiling_face:

It will make maintenance difficult and fewer developers will understand and contribute to the code.

devemux86:
@menion 

I merged it via 4bb9457cebc0a931f53bcdb5430b7d467db6fcda with the [above restore](https://github.com/mapsforge/mapsforge/pull/1741#discussion_r2375578442) of previous code for same rendering.
So now we just have the important performance improvement with the `tileAsGeometry`.

Thanks again for the contribution! :+1:

PR #1795, opened 2026-05-04
Remove AndroidMaps URL from map providers list

kylecorry31:
The AndroidMaps URL was redirecting to ads rather than offering map downloads.

Addresses #1796


devemux86:
Thanks, map providers checked for updates with https://github.com/mapsforge/mapsforge/pull/1797.

PR #1799, opened 2026-05-09
Update Mapsforge-Applications.md to include Trail Sense

kylecorry31:
Trail Sense will be adding a map layer for displaying Mapsforge maps in version 7.8.0 (releasing toward the end of May / early June). I'm using the Mapsforge library to generate tile bitmaps which are loaded by the existing (custom) map view.


devemux86:
Thanks!

PR #1805, opened 2026-07-21
Update all dependencies to latest versions

Sublimis:
How about...?


devemux86:
Too many changes in 1 pull request that doesn't even build successfully.
All of these need to be discussed separately first.

**Important: we want to be compatible with IDEA to cover most users.**

IDEA 2026.2 currently has some problems with the Android plugin.

IDEA 2026.1 works with Android Gradle plugin 9.0.1 which needs Gradle 9.1.0:
https://developer.android.com/build/releases/agp-9-0-0-release-notes

Also many of the other updates are not really needed.

When IDEA moves forward successfully,
we will consider advancing Gradle after thorough testing and depending on the version that works each time.

